PR #56, opened 2022-06-18
Cleanup compilation warnings

alosarjos:
Cleaning up some compilation warnings for new features code not being under corresponding CFG flag, and an unused  `use` I saw while testing the package.

PD: 0.9 is published in AUR ;)

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -9,7 +9,7 @@ use crate::{
 };
 
 #[cfg(any(feature = "lyric-finder", feature = "image"))]
-use anyhow::Context as AnyHowContext;
+use anyhow::Context as _;
 use anyhow::Result;
 use librespot_core::session::Session;
 use rspotify::prelude::*;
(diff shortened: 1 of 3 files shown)


aome510:
Thanks @alosarjos. LGTM generally. Let's wait for the checks to make sure we don't miss anything.

alosarjos:
By bad, this AnyHowContext needs to be in the Scope if any of two different features is enabled. I think it should be fine now.

spotify_player/src/client/mod.rs
```diff
@@ -7,7 +7,10 @@ use crate::{
     event::{ClientRequest, PlayerRequest},
     state::*,
 };
-use anyhow::{Context as AnyhowContext, Result};
+
+#[cfg(any(feature = "lyric-finder", feature = "image"))]
+use anyhow::Context as AnyHowContext;
```
aome510, line 12:
```suggestion
use anyhow::Context as _;
```

I just realize we don't really to name the import here.

alosarjos: Yeah, I think it's only a matter of getting it in the scope. I guess there is no need to rename it.

PR #113, opened 2023-01-27
Add support for seeing the current queue

BeneCollyridam:
I don't know if it would be of interest, but for my own use I implemented a
popup showing the current queue.

Feel free to close the PR or give me feedback to get it to a state, where you
think it can be merged

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -201,12 +201,11 @@ impl Client {
             #[cfg(feature = "lyric-finder")]
             ClientRequest::GetLyric { track, artists } => {
                 let client = lyric_finder::Client::from_http_client(&self.http);
-                let query = format!("{} {}", track, artists);
+                let query = format!("{track} {artists}");
 
                 if !state.data.read().caches.lyrics.contains(&query) {
                     let result = client.get_lyric(&query).await.context(format!(
-                        "failed to get lyric for track {} - artists {}",
-                        track, artists
+                        "failed to get lyric for track {track} - artists {artists}"
                     ))?;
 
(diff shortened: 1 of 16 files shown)


spotify_player/src/client/mod.rs
```diff
@@ -345,6 +345,13 @@ impl Client {
             ClientRequest::DeleteFromLibrary(id) => {
                 self.delete_from_library(state, id).await?;
             }
+            ClientRequest::GetCurrentUserQueue => {
+                tracing::info!("Fetching current user queue");
```
aome510, line 349: Nit: no need to add a log here. This kind of logging should already be handled by the caller function.

spotify_player/src/ui/popup.rs
```diff
@@ -280,3 +289,63 @@ pub fn render_commands_help_popup(
...
+                .map(|(i, x)| {
+                    Row::new(vec![
+                        Cell::from(format!("{}", i + 1)),
+                        Cell::from(get_playable_name(x).to_string()),
+                    ])
+                })
+                .collect::<Vec<_>>(),
+        )
+        .header(
+            Row::new(vec![Cell::from("No"), Cell::from("Title")]).style(ui.theme.table_header()),
```
aome510, line 341: Nit: for consistency with track table, "No" should be changed to "#"

aome510 (commented):
Thanks for the PR. Looks good to me. Only a few nits which I can update by myself

aome510:
It would be nice for the queue to be automatically updated by listening to playback events. This would be a non-trivial implementation, so I'll leave it to a follow-up PR/separate issue.

PR #118, opened 2023-02-03
More flexible theme config

daschw:
Hi, thanks for this awesome piece of software!

I just recently came across this project and played around with the theme config a little. At some point I thought it would be nice to have darker borders. However, I could not find this option anywhere in the component styles. So I thought I'd try to implement this myself. After I had finished, I realized that this would require every user to update their theme.toml files because without the new `border` option spotify_player would not be able to parse them and error. Furthermore I saw that you recently refactored the theming config and deleted `fg_selected` and `bg_selected` (or how they were called) and I saw the TODO comment to find a solution to not having to set each component style when adding this section in theme.toml.

Hence I thought, before I add an option to change the border I'll try to figure out a way to solve this TODO. Afterwards, adding new component styles would not be an issue, because it would not affect older configs, that did not set this attribute. So that is what I ended up doing. With this PR
- it is possible to only set some of the component styles in theme.toml
- it is possible to not set the palette in theme.toml (and just modify component styles)
- it is possible to set the border style with the new component_style attribute `border`
- it is possible to set the style of the current selection with the new component_style `selection`

I am quite sure that none of these changes affect any existing theme configs.

With this I could now have a theme.toml like this
```toml
[[themes]]
name = "mine"
[themes.component_style]
border = { fg = "BrightBlack" }
selection = { fg = "Yellow", modifiers = ["Bold"] }
```
that would look like this with my terminal colors:

Screenshot_20230203_203849

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -136,6 +136,7 @@ If specified, a field's value must be set to be a hex representation of a RGB co
 To define application's component styles, the user can specify any of the below fields:
 
 - `block_title`
+- `border`
 - `playback_track`
 - `playback_album`
 - `playback_metadata`
@@ -150,6 +151,7 @@ Default value for application's component styles:
 
 ```toml
 block_title = { fg = "Magenta"  }
+border = {} # uses the palette's fg and bg color
 playback_track = { fg = "Cyan", modifiers = ["Bold"] }
 playback_album = { fg = "Yellow" }
 playback_metadata = { fg = "BrightBlack" }
--- a/docs/config.md
+++ b/docs/config.md
@@ -133,7 +133,7 @@ If specified, a field's value must be set to be a hex representation of a RGB co
 
 ### Component Styles
 
-To define application's component styles, user needs to specify **all of the below fields**:
+To define application's component styles, the user can specify any of the below fields:
 
 - `block_title`
 - `playback_track`
--- a/docs/config.md
+++ b/docs/config.md
@@ -144,6 +144,7 @@ To define application's component styles, the user can specify any of the below 
 - `current_playing`
 - `page_desc`
 - `table_header`
+- `selection`
 
 A field in the component styles is a `Style` struct which has three optional fields: `fg`, `bg` and `modifiers`. `fg` and `bg` can be either a palette's color (string in pascal case) or a custom RGB color using the following format: `fg = { Rgb { r = ..., g = ..., b = ... } }`. `modifiers` can only be either `Italic` or `Bold`.
 
@@ -159,6 +160,7 @@ playback_progress_bar = { bg = "BrightBlack", fg = "Green" }
 current_playing = { fg = "Green", modifiers = ["Bold"] }
 page_desc = { fg = "Cyan", modifiers = ["Bold"] }
 table_header = { fg = "Blue" }
+selection = {} # if not set, this is bold and reverses fg and bg color
 ```
 
 ## Keymaps
--- a/docs/config.md
+++ b/docs/config.md
@@ -89,7 +89,7 @@ The application's theme can be modified by setting the `theme` option in `app.to
 
 A theme has three main components: `name` (the theme's name), `palette` (the theme's color palette), `component_style` (a list of pre-defined styles for application's components).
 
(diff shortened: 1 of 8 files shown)


aome510:
Hi @daschw, thanks for your interests in this project and the effort you put in making this PR.

I'll take a look and review this PR the next few days.

docs/config.md
```diff
@@ -89,7 +89,7 @@ The application's theme can be modified by setting the `theme` option in `app.to
 
 A theme has three main components: `name` (the theme's name), `palette` (the theme's color palette), `component_style` (a list of pre-defined styles for application's components).
 
-`name` and `palette` are required when defining a new theme. If `component_style` is not specified, a default value will be used.
+`name` is required when defining a new theme. If `palette` is not set, the terminals colorscheme will be used. If `component_style` is not specified, a default value will be used.
```
aome510, line 92:
```suggestion
`name` is required when defining a new theme. If `palette` is not set, a palette based on the terminal's colorscheme will be used. If `component_style` is not specified, a default value will be used.
```

docs/config.md
```diff
@@ -133,30 +133,34 @@ If specified, a field's value must be set to be a hex representation of a RGB co
...
 - `table_header`
+- `selection`
 
 A field in the component styles is a `Style` struct which has three optional fields: `fg`, `bg` and `modifiers`. `fg` and `bg` can be either a palette's color (string in pascal case) or a custom RGB color using the following format: `fg = { Rgb { r = ..., g = ..., b = ... } }`. `modifiers` can only be either `Italic` or `Bold`.
 
 Default value for application's component styles:
 
 ```toml
 block_title = { fg = "Magenta"  }
+border = {} # uses the palette's fg and bg color
```
aome510, line 155: Instead of putting a comment here, can you add a note after component style's document that mentions that the default values for `fg` and `bg` are palette's `fg` and `bg`?

docs/config.md
```diff
@@ -133,30 +133,34 @@ If specified, a field's value must be set to be a hex representation of a RGB co
...
 playback_track = { fg = "Cyan", modifiers = ["Bold"] }
 playback_album = { fg = "Yellow" }
 playback_metadata = { fg = "BrightBlack" }
 playback_progress_bar = { bg = "BrightBlack", fg = "Green" }
 current_playing = { fg = "Green", modifiers = ["Bold"] }
 page_desc = { fg = "Cyan", modifiers = ["Bold"] }
 table_header = { fg = "Blue" }
+selection = {} # if not set, this is bold and reverses fg and bg color
```
aome510, line 163:
```suggestion
selection = { modifiers = ["Bold", "Reverse"] } 
```

Note that this requires you to add a new enum variant `StyleModifier::Reverse`

aome510 (changes requested):
This PR is easier to review than I initially thought. Generally, it LGTM. Only a few small changes are needed.

aome510:
> ```toml
> [[themes]]
> name = "mine"
> [themes.component_style]
> border = { fg = "BrightBlack" }
> selection = { fg = "Yellow", modifiers = ["Bold"] }
> ```

Tried your theme and it also works for me!

daschw:
Hi, thanks for the review! I implemented all your suggestion. I was not sure, where exactly to put the default 'fg' and 'bg' value description instead of the "border comment". Let me know, if this is not what you had in mind.

docs/config.md
daschw: implemented in https://github.com/aome510/spotify-player/pull/118/commits/591568db595cbe27b517b96e07f6837008b97272

docs/config.md
daschw: implemented in https://github.com/aome510/spotify-player/pull/118/commits/591568db595cbe27b517b96e07f6837008b97272

docs/config.md
daschw: implemented in https://github.com/aome510/spotify-player/pull/118/commits/591568db595cbe27b517b96e07f6837008b97272 - not sure about the location of the comment, hope, that's OK

aome510 (approved):
Thanks for the changes. LGTM!

aome510:
I also update the example theme and `theme_parse` script by myself.

Issue #111, opened 2023-01-25
API for consume/dispatch actions from third-party applications


diegoulloao:
**Is your feature request related to a problem? Please describe.**
Not a problem. is possible to dispose of a simple API to consume in order to dispatch actions from third-party applications, actions like:
- get current song title that's playing
- get the current playlist that's playing
- play a specific song of the current playlist
- go to next/previous song of the current playlist

**Describe the solution you'd like**
This would work sending commands to any existent instance of spotify-player running.
I would like to write a neovim plugin to control spotify-player while coding, as a widget of spotify-player for neovim, but it needs an API exposed.

E.g.
```bash
spotify_player --cmd="get-playlist"
spotify_player --cmd="play-song id"
```

I'm not sure if I can do this by using the service that spotify_player is using in background.
I don't know if that will get sync with spotify_player either (without hit an API for auth, etc)

I would like to get this idea working hand to hand with spotify-player, because I think is an app with a great future ahead.

Sounds crazy?

aome510:
Hi @diegoulloao, thanks for creating this issue. Actually, I have a plan to create a similar one as I've got some related requests regarding external APIs to communicate with the app.

Related issues:
- https://github.com/aome510/spotify-player/issues/103

> Sounds crazy?

I thought about implementing the same thing before. I don't think it's impossible. I'll try and see xD.

diegoulloao:
Yes, excellent. LMK if you try something, you can start with simple actions to test. In parallel I can start building the bases for the neovim/plugin widget if something comes out.

PR #137, opened 2023-02-17
Implement notification support

### LINKED ISSUE #123: feature request: notifications for tracks
The official Spotify client (on linux) sends notifications when a new track is played, or when you reach the end of a song, and the next one starts playing. It would be great if this could be added to `spotify-player` :point_right: :point_left: 

Here's an example of how it looks like with the official client:
image

aome510:
Resolves #123 

## Changes

- added `notify` feature (with `notify_rust` as dependencies) for desktop notification support
- saved downloaded cover images into `$APP_CACHE_FOLDER/image`


apprehensions:
seems to work, but it seems that the format of the notification is hardset, and i would suggest it could be modified (although, the current is pretty good!)

it seems that the D-Bus `PlayBackStatus` when spotify_player is playing a song is constantly being set, and i would suggest it be sent once; as i've never seen the D-Bus requests like this on any music player (including Spotify itself)

as for the cache, it technically works, but the way to get the 'latest' album cover is via linux's `ctime` (creation time) or parsing the current song and formatting (not possible in spotify_player yet) it to the appropriate file name.

apprehensions:
another feature i find cool is that the icon of the current playing song is passed to the D-Bus notification from the cache image file rather than encoded data (as seen in most applications; the output from `dbus-monitor` when an image is encoded in the request is 5171 lines long.).

aome510:
> seems to work, but it seems that the format of the notification is hardset, and i would suggest it could be modified (although, the current is pretty good!)

Yeah, making the notification format configurable is not in the scope of this PR, maybe a follow-up PR. Do u have any suggestions on another format? Currently the format is "{name} • {artists}" as the summary and "{album}" as the body.

> it seems that the D-Bus PlayBackStatus when spotify_player is playing a song is constantly being set, and i would suggest it be sent once; as i've never seen the D-Bus requests like this on any music player (including Spotify itself)

Does it happen outside this PR or only with this PR? Also, not too familiar with D-BUS, how did you know that it's constantly being set?

> as for the cache, it technically works, but the way to get the 'latest' album cover is via linux's ctime (creation time) or parsing the current song and formatting (not possible in spotify_player yet) it to the appropriate file name.

I have an exactly same thought. `ctime` doesn't really work if you play cached track the second time. So I guess we should wait for #111.

> another feature i find cool is that the icon of the current playing song is passed to the D-Bus notification from the cache image file rather than encoded data (as seen in most applications; the output from dbus-monitor when an image is encoded in the request is 5171 lines long.).

I'm not sure what you mean by this. Do you mind to elaborate further?

apprehensions:
> Does it happen outside this PR or only with this PR? Also, not too familiar with D-BUS, how did you know that it's constantly being set?

`dbus-monitor`

> I'm not sure what you mean by this. Do you mind to elaborate further?

don't worry about it, it was just a compliment.

aome510:
> actually, happens with vanilla as well.

Interesting, will take a look.

> don't worry about it, it was just a compliment

Oh ok, ty <3. I initially thought you requested for a new feature

aome510:
image

aome510:
@apprehensions regarding the `PlaybackStatus` change thingy, this happened because `PlaybackStatus` is coupled with the playback's progress for reporting `position` and the status change is always emitted without checking the previous value as in https://github.com/Sinono3/souvlaki/blob/3b5d66fc4035b7a1e7ffa899a1636a3ce10a2571/src/platform/mpris/dbus.rs#L429-L433

So there's nothing much I can do with `spotify-player`. I would suggest raising issue or making a PR in the `souvlaki` repo if this really annoys you.

PR #151, opened 2023-03-02
Windows duplicate keys

### LINKED ISSUE #136: [Windows] Every keypress is duplicated
**Describe the bug**
On windows, every keypress is sent both on the press down, and on the key release. This makes the app near impossible to use as searches can't be made and only even rows can be selected

**To Reproduce**
`cargo install spotify_player` on Windows 11
`spotify_player.exe` (Issue occurs in both cmd.exe and powershell)
Navigate up and down, try searching, etc

**Expected behavior**
A key is sent to the app once per keypress.

**Log and backtrace**
Reference log and backtrace, which are usually located in `$HOME/.cache/spotify-player/`, of a run reproducing the bug.
'spotify-player-23-02-16-17' in the cache directory seems to be an empty file even after using the program

**Screenshots**
https://user-images.githubusercontent.com/42524580/219504557-0267b5c6-52ea-441b-b5d5-f29fc40ed41d.mp4

**Environment**

- OS: Windows 11 build 25295
- Application version: 0.11.1
- Application features: if applicable, provide [cargo features](https://doc.rust-lang.org/cargo/reference/features.html) used to build with `spotify-player`: Defaults (Installed with `cargo install spotify_player`

**Additional context**
Doesn't appear to happen on my Linux install.

ViniciusAndr:
Resolves #136 

Hi, I saw that the error happened because the code was called twice, I couldn't figure out the reason, but this change worked for me.

I put a timer to give a delay of 0.3 seconds for each command only when it is windows and that solved the error, as I have no experience in rust so I don't know if there would be another way to solve it

### DIFF
--- a/spotify_player/src/event/mod.rs
+++ b/spotify_player/src/event/mod.rs
@@ -8,6 +8,8 @@ use crate::{
 #[cfg(feature = "lyric-finder")]
 use crate::utils::map_join;
 use anyhow::Result;
+use std::time::{Duration, Instant};
+use std::env;
 
 mod page;
 mod popup;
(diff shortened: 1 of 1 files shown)


aome510:
> Hi, I saw that the error happened because the code was called twice, I couldn't figure out the reason, but this change worked for me.

Hi, thanks for raising this. I actually did a quick search on https://github.com/crossterm-rs/crossterm and found https://github.com/crossterm-rs/crossterm/issues/752. I kinda know how to fix this issue rn, will update in this PR.

aome510:
@ViniciusAndr can you pull the new change and try to see if it actually solves the issue?

ViniciusAndr:
@aome510 Ok I will test

ViniciusAndr:
@aome510 Your commit solved the issue

PR #168, opened 2023-03-25
remove colon from log file name

### LINKED ISSUE #167: Windows Logging not working
**Describe the bug**
Logging does not work on Windows. A log file is created but the file name is truncated like 'spotify-player-23-03-25-14' rather than 'spotify-player-23-03-25-14:xx.log' and nothing is written to it.

**To Reproduce**
Run spotify-player. Nothing is written to the log file during or after execution.

felixkroemer:
Fixes #167 

Windows file names can't have colons. I changed the log file format accordingly. If you want to keep the colon you can use ꞉ as described [here](https://stackoverflow.com/questions/10386344/how-to-get-a-file-in-windows-with-a-colon-in-the-filename).

### DIFF
--- a/spotify_player/src/main.rs
+++ b/spotify_player/src/main.rs
@@ -84,7 +84,7 @@ async fn init_spotify(
 fn init_logging(cache_folder: &std::path::Path) -> Result<()> {
     let log_prefix = format!(
         "spotify-player-{}",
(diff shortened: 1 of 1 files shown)


aome510 (approved):
LGTM! Thanks for the fix

PR #221, opened 2023-07-17
Create app.toml with default values if it doesn't exist

### LINKED ISSUE #182: Create app.toml with default values if it doesn't exist
I was trying to figure out how to enable media controls on macOS. It took me a while to figure out the correct syntax of app.toml by looking in `examples` directory.

If `app.toml` doesn't exist in `$HOME/.config/spotify-player`, create it with default values, so it's easier for users to modify it.

Do nothing.

**Additional context**
Thanks for creating spotify-player. It seems really solid.

cobbinma:
If an application configuration file is not found, one will be created with default values in the config directory.

The configs have been refactored to use `new` constructors instead of default implementations.

closes #182

### DIFF
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -229,7 +229,7 @@ impl Default for DeviceConfig {
 impl AppConfig {
     pub fn new(path: &Path, theme: Option<&String>) -> Result<Self> {
         let mut config = Self::default();
-        if config.parse_config_file(path)?.is_none() {
+        if !config.parse_config_file(path)? {
             config.write_config_file(path)?
         }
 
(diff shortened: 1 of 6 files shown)


spotify_player/src/state/mod.rs
```diff
@@ -10,7 +10,7 @@ pub use model::*;
 pub use player::*;
 pub use ui::*;
 
-use crate::config;
+use crate::config::{self};
```
aome510, line 13: This is a weird import. Is it different from `use crate::config`?

spotify_player/src/state/mod.rs
```diff
@@ -33,40 +33,23 @@ pub struct State {
 
 impl State {
     /// parses application's configurations
```
aome510, line 35:
```suggestion
    /// creates an application's state based on files in a configuration folder and an optional pre-defined theme
```

spotify_player/src/config/mod.rs
```diff
@@ -227,14 +227,38 @@ impl Default for DeviceConfig {
...
+            config.theme = theme.to_owned()
+        }
+
+        Ok(config)
+    }
+
     // parses configurations from an application config file in `path` folder,
     // then updates the current configurations accordingly.
-    pub fn parse_config_file(&mut self, path: &Path) -> Result<()> {
+    fn parse_config_file(&mut self, path: &Path) -> Result<Option<()>> {
```
aome510, line 245: I would prefer returning `Result<bool>` and the function document should be updated with: "Returns false if no config file found and true otherwise".

aome510 (changes requested):
Thanks for looking into this. Look good! Only need a few changes.

spotify_player/src/state/mod.rs
cobbinma: Think this happened as I imported AppConfig directly at some stage e.g. `use crate::config::{self, AppConfig}`.

Reverted anyway 👍

aome510 (approved):
ty!

PR #222, opened 2023-07-23
Playlist CLI Editing - New, Delete, Import, Fork, Update

justjokiing:
Added `playlist` CLI commands:

```
spotify_player playlist -h

Playlist editing

Usage: spotify_player playlist <COMMAND>

Commands:
  new     Create a new playlist
  delete  Delete a playlist
  import  Imports all songs from a playlist into another playlist.
  list    Lists all user playlists.
  fork    Creates a copy of a playlist and imports it.
  sync    Syncs imports for all playlists or a single playlist.
  help    Print this message or the help of the given subcommand(s)

Options:
  -h, --help  Print help
```


spotify_player/src/cli/mod.rs
```diff
@@ -59,15 +59,41 @@ pub enum IdOrName {
...
+    PlaylistImport {
+        from: PlaylistId<'static>,
+        to: PlaylistId<'static>,
+    },
+    PlaylistFork {
+        id: PlaylistId<'static>,
+    },
+    PlaylistUpdate {
+        id: Option<PlaylistId<'static>>,
+    },
```
aome510, line 96: I think it should be put into another enum `PlaylistCommand`. The initial enum can be renamed to `PlaybackCommand`.

aome510: With that, the variant's name can be reduced, for example, `PlaylistNew` to `New` for better clarity.

spotify_player/src/cli/handlers.rs
```diff
@@ -178,8 +179,94 @@ pub fn handle_cli_subcommand(
...
+    let (cmd, args) = args.subcommand().expect("playlist subcommand is required");
+    let command = match cmd {
+        "new" => {
+            let empty = String::new();
+
+            let name = args
+                .get_one::<String>("name")
+                .expect("name arg is required");
+
+            let description = args.get_one::<String>("description").or(Some(&empty));
```
aome510, line 198: instead of making `description` `Option<&String>` then unwrapping it later, can initialize it as a `String` instead:

```rust
let description = args.get_one::<String>("description").map(|s| s.to_owned()).unwrap_or_default();
```

spotify_player/src/cli/commands.rs
```diff
@@ -128,3 +128,43 @@ pub fn init_like_command() -> Command {
...
+            .arg(Arg::new("from")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new()))
+            .arg(Arg::new("to")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new()))
+            .after_help("Copies all tracks into a playlist. If imported again, it adds new tracks and outputs information on deleted tracks since last import."))
+        .subcommand(Command::new("list").about("Lists all user playlists."))
+        .subcommand(Command::new("fork").about("Creates a copy of a playlist and imports it.")
+            .arg(Arg::new("id")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new())))
+        .subcommand(Command::new("update").about("Updates all imported playlists or a single individual import.")
```
aome510, line 166: Haven't read the implementation but based on the command's description, I don't understand the purpose of this command.

spotify_player/src/cli/client.rs
```diff
@@ -130,6 +141,16 @@ async fn handle_socket_request(
...
             Ok(Vec::new())
         }
+        Request::Playlist(command) => {
+            let resp = handle_playlist_request(client, state, command).await?;
+
+            // Allowing command output of success
+            let mut resp_vec = serde_json::to_vec_pretty(&resp)?;
+            // Removing quotation marks
+            resp_vec.pop();
+            resp_vec.remove(0);
```
aome510, line 151: The `resp` is a string, why don't use `String::into_bytes` instead of converting from  `json` and manually removing the quotation marks?

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
             };
             PlayerRequest::SeekTrack(progress + chrono::Duration::milliseconds(position_offset_ms))
         }
+        _ => unreachable!(),
```
aome510, line 368: I don't think `unreachable!` is needed here. `unreachable!` should be avoided when matching `enum` as it defeats the purpose of enum exhaustive checking.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+        Command::PlaylistDelete { id } => {
+            let user = state.data.read().user_data.user.to_owned().unwrap();
+            let uid = user.id;
+
+            let playlist_check = client
+                .spotify
+                .playlist_check_follow(id.to_owned(), &[uid])
+                .await;
+            if playlist_check.is_err() {
+                anyhow::bail!("Could not find playlist {}", id.to_string())
```
aome510, line 416: `to_string` shouldn't be needed as it also means `id` implements `fmt::Display`.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+                    name.as_str(),
+                    Some(public),
+                    Some(collab),
+                    Some(description.as_str()),
+                )
+                .await?;
+            Ok(format!(
+                "Playlist '{}' with id '{}' was created.",
+                resp.name,
+                resp.id.to_string()
```
aome510, line 404: `to_string` is not needed for `id`

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+                .spotify
+                .playlist_check_follow(id.to_owned(), &[uid])
+                .await;
+            if playlist_check.is_err() {
+                anyhow::bail!("Could not find playlist {}", id.to_string())
+            }
+
+            // Won't delete if not following
+            let following = playlist_check.unwrap().pop();
+            if following.unwrap() {
```
aome510, line 421:
```suggestion
            let following = playlist_check.unwrap().pop().unwrap();
            if following {
```

can move the `uwnrap` upper.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+                client.spotify.playlist_unfollow(id.to_owned()).await?;
+                Ok(format!("'{}' was deleted", id.to_string()))
+            } else {
+                Ok(format!("'{}' was not followed.", id.to_string()))
```
aome510, line 425:
```suggestion
                Ok(format!("Playlist '{id}' was deleted/unfollowed.")
            } else {
                Ok(format!("Playlist '{id}' is not followed by the user, nothing to be done.")
```

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+            }
+        }
+        Command::PlaylistList => {
+            let resp = client
+                .spotify
+                .current_user_playlists_manual(Some(50), None)
+                .await?;
```
aome510, line 432: can use `client.current_user_playlists()` instead as it does page fetching automatically.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+
+            let mut out = String::new();
+            for pl in resp.items {
+                // Opposite way, don't know which is better
+                // Might want to add color
+                //out.push_str(format!("{}:{}\n", pl.name, pl.id.id()).as_str());
```
aome510, line 438: I think `id: name` is better.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+        Command::PlaylistFork { id } => {
+            let user = state.data.read().user_data.user.to_owned().unwrap();
+            let uid = user.id;
+
+            let from = client
+                .spotify
+                .playlist(
+                    id.to_owned(),
+                    None,
+                    Some(Market::Country(Country::UnitedStates)),
```
aome510, line 458: Not everyone is from `US` should be `None` or at least `Market::FromToken` instead.

spotify_player/src/cli/client.rs
```diff
@@ -344,9 +365,435 @@ async fn handle_playback_request(
...
+                }
+
+                Ok(result)
+            }
+        }
+        _ => unreachable!(),
+    }
+}
+
+async fn playlist_import(
```
aome510, line 587: can you add a function documentation describing what this function does?

aome510 (changes requested):
Still cannot wrap my head around the idea of playlist importing in this PR, like why does it need to access to the filesystem?

Will come back to this later next week. In the meantime, you can try to address my current comments.

spotify_player/src/cli/client.rs
justjokiing: I agree, but I don't think it will compile without. Doesn't each 'handle' only match part of the Command enum?

spotify_player/src/cli/commands.rs
justjokiing: The update command is used to update imports. This just means that any imported playlists will be re imported. It is for ease of use. It would be more difficult for me to find out what was imported and do a command for each.

spotify_player/src/cli/client.rs
justjokiing: Nevermind, implemented the PlaylistCommand enum so now it doesn't need it.

justjokiing:
> Still cannot wrap my head around the idea of playlist importing in this PR, like why does it need to access to the filesystem?

My current implementation writes to an 'import' folder with the ids of the imported-to playlist as the dir and a file with the imported playlist information.

Having this file allows us to keep track of what as already been imported. Then when imported again (like the update command), allows us to quickly check if there are any changes to the imported playlist. Then if there are changes, we can see what has changed since the last import.

justjokiing:
I am having an issue with my import though. I cannot seem to find a way to get more than 100 of the playlist's tracks.

aome510:
> I am having an issue with my import though. I cannot seem to find a way to get more than 100 of the playlist's tracks.

The maximum tracks in a playlist returned by `client.spotify.playlist(...)` is `50` iirc. You can use `client.playlist_context(...)` instead or you can look into the function's implementation for reference on how to use `all_paging_items`:

https://github.com/aome510/spotify-player/blob/5aa06ba5ad3abbf901b36fdec13c94f3a48916c3/spotify_player/src/client/mod.rs#L989-L1013

`cargo format` CI keeps falling (https://github.com/aome510/spotify-player/actions/runs/5647354940/job/15346517162?pr=222). Do you enable auto-formatting with `rustfmt` in your code editor? You can also enable a linter (`cargo clippy`), so that it can catch some bad coding styles.

aome510:
> My current implementation writes to an 'import' folder with the ids of the imported-to playlist as the dir and a file with the imported playlist information.
> 
> Having this file allows us to keep track of what as already been imported. Then when imported again (like the update command), allows us to quickly check if there are any changes to the imported playlist. Then if there are changes, we can see what has changed since the last import.

Thanks for the clarification.

Why do you need to write to files though? You can do the import in memory. I don't think storing states into files helps that much with the current implementation.

spotify_player/src/cli/commands.rs
aome510: Iiuc, for each playlist, you want to store a list of "from" playlists imported to that playlist. `update` command basically runs `import` with that playlist and each of the "from" playlists from the list.

tbh, I'm not too fond of the idea of having this as a CLI command as I don't find a lot of use cases for this feature.  You can create an external script to do that with the `import` command.

spotify_player/src/cli/client.rs
aome510: Note: this is not resolved yet

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,425 @@ async fn handle_playback_request(
...
+
+                writeln!(file, "{}:{}", id.id(), track.name)?;
+
+                // Add tracks to hashset
+                new_ids.insert(track);
+
+                if !to_hash.contains(&id) {
+                    client
+                        .spotify
+                        .playlist_add_items(import_to.to_owned(), [PlayableId::Track(id)], None)
```
aome510, line 718: `playlist_add_items` receives a list of ids. So instead of making a single for each id, you can determine the intersection list between `from` playlist and `to` playlist and call `playlist_add_items` with ids from that list.

As you already use `HashSet` to store the track's ids. You can use the `differences` method (https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.difference).

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,425 @@ async fn handle_playback_request(
...
+}
+
+/// Imports a playlist into another playlist.
+/// All tracks from the imported playlist are added to the import-to playlist if they are not in there already.
+/// The state of the imported playlist and a hash of the track ids are written to a file
+///
+/// This file is used in subsequent imports of the same two playlists
+/// It is first used to see if the state of the imported playlist has changed
+/// then if so, is used to see what has differed in the change.
+async fn playlist_import(
```
aome510, line 577: Assuming that importing a `from` playlist to a `to` playlist means adding tracks that belong to `from` but don't belong to `to`, this function's implementation shouldn't be more than 20-30 lines. 

IMHO, this is kinda overcomplicated.

aome510:
Also, forget to say that thanks for the PR. 

Hope that my suggestions don't sound harsh to you. I just want the codes to be easy to read and self-explanatory to reduce the cost of future maintenance.

spotify_player/src/cli/client.rs
justjokiing: I am working on an implementation that uses a buffer of under 100 to send through `playlist_add_items`, just doing the difference may cause a payload larger than the API allows

spotify_player/src/cli/commands.rs
justjokiing: Without the update command, managing imports would be tedious. It could definitely be a script but why should a user have to set up a script to keep their playlists imported?

I will and do use the update command the most. Imports/forks are done once but keeping them updated will be an everyday thing.

spotify_player/src/cli/client.rs
justjokiing: Fixed, will be in next commit

justjokiing:
> `cargo format` CI keeps falling (https://github.com/aome510/spotify-player/actions/runs/5647354940/job/15346517162?pr=222). Do you enable auto-formatting with `rustfmt` in your code editor? You can also enable a linter (`cargo clippy`), so that it can catch some bad coding styles.

I use Neovim and have just been forgetting to run `cargo fmt` before commits. I will make sure to use that and clippy before the next commit.

spotify_player/src/cli/client.rs
justjokiing: I humbly disagree. It is slightly more than what it could be, but its slight increase in complexity reduces extra iterations and allows for important features. 

The logic follows 3 possible states:
- New import - add all tracks and write to file
- Old import, no change - check that hash is the same, then do nothing
- Old import, changes - add new tracks and write to file

Would there be a way for me to break this down to make it more simple and readable?

justjokiing: > Assuming that importing a `from` playlist to a `to` playlist means adding tracks that belong to `from` but don't belong to `to`, this function's implementation shouldn't be more than 20-30 lines.

Another reason I forgot to mention is that I do not want to just add anything that is not in `to` but any changes in `from` into `to`. The difference is that I do not add anything that was already added before, meaning you can delete items that were added from an import and the subsequent imports will not add them back. This is a desired action.

For this to work, I cannot just request to see the state of `from`, but look to the file I wrote to see what has been added before.

I use this feature specifically. I am importing a friends playlist that I like but I don't like some of the songs, so I delete them. The deleted songs do not get re-added  but I get any new additions to my friends playlist.

aome510:
Thanks @justjokiing. I'll come back to this PR this weekend hopefully.

spotify_player/src/cli/client.rs
aome510: > Another reason I forgot to mention is that I do not want to just add anything that is not in `to` but any changes in `from` into `to`. The difference is that I do not add anything that was already added before, meaning you can delete items that were added from an import and the subsequent imports will not add them back. This is a desired action.
> 
> For this to work, I cannot just request to see the state of `from`, but look to the file I wrote to see what has been added before.
> 
> I use this feature specifically. I am importing a friends playlist that I like but I don't like some of the songs, so I delete them. The deleted songs do not get re-added but I get any new additions to my friends playlist.

Ah, now it makes more sense why you need to file logs. 

My point is that this adds another layer to the things you need to know when using the CLI commands. When looking at the `import`, I'll expect it to do one thing: importing all tracks from one playlist to another playlist without duplications.

From a user standpoint, when looking into the command description without looking into the implementation, I have no idea why there is `delete` option for `import` command and the meaning of the `update` command. That means you'll need to explain to users the concept of storing playlist import states into files and how we do it for users to be able to use those commands.

I mean it still comes down to a matter of personal preferences. I want the CLIs to be simple and just do one thing by default. You can build more complicated behavior by combining different CLI usages, hence writing scripts.

aome510: Let me review this PR again to see if the implementation can be simplified and if it's possible to improve the documentation so that users don't need to know about the underlying import logic with the filesystem.

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+    client: &Client,
+    import_from: PlaylistId<'static>,
+    import_to: PlaylistId<'static>,
+    delete: bool,
+) -> Result<String> {
+    #[derive(PartialEq, Eq, Hash)]
+    struct PlaylistData {
```
aome510, line 584: Should this be `TrackData` instead?

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+            name: track.name,
+        };
+
+        from_ids.push(song);
+        hasher.write(track.id.id().as_bytes());
+    }
+    let hash = hasher.finish().to_string();
+
+    // Get import dir/file
+    let conf_dir = get_config_folder_path()?;
```
aome510, line 625: should use a cache folder instead of a config folder.

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+                writeln!(file, "{}:{}", id.id(), track.name)?;
+
+                // Add tracks to hashset
+                new_ids.insert(track);
+
+                if !to_hash.contains(&id) {
+                    client
+                        .spotify
+                        .playlist_add_items(import_to.to_owned(), [PlayableId::Track(id)], None)
+                        .await?;
```
aome510, line 716: You don't need the import single-item logic any more right?

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+
+    // Create hashset of import_to playlist
+    // Will use to check before adding track
+    let mut to_hash = HashSet::new();
+    for track in to_tracks {
+        to_hash.insert(track.id);
+    }
+
+    // Create hash and vec of 'from_ids'
+    let mut hasher = DefaultHasher::new();
```
aome510, line 611: There is a problem with using a hash to compare changes as this assumes static ordering between items in a playlist. For example, a hash of `A, B, C` will be different than `C, A, B`.

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+                }
+            }
+
+            if delete {
+                result.push_str("These tracks have been deleted from the playlist.");
+                client
+                    .spotify
+                    .playlist_remove_all_occurrences_of_items(to.id, delete_buff, None)
+                    .await?;
+            }
```
aome510, line 803: I see, so the `delete` option is to remove **all** tracks from the previous version of the playlist but not from the current version. Should make this more clear in the command's document instead of

```
.help("Will delete any tracks that the import does."))
.after_help("Copies all tracks into a playlist. If imported again, it adds new tracks and outputs information on deleted tracks since last import.")
```

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +365,453 @@ async fn handle_playback_request(
...
+                .create(true)
+                .open(&from_file)?;
+
+            writeln!(file, "{}", hash)?;
+            writeln!(file)?;
+
+            for track in from_ids {
+                let id = track.id.to_owned();
+
+                writeln!(file, "{}:{}", id.id(), track.name)?;
```
aome510, line 707: If you update the file cache by just storing the latest playlist's data. I don't think it helps with the "don't add deleted tracks" use case you mentioned in https://github.com/aome510/spotify-player/pull/222/files#r1277490271.

For example, let's say you initially import tracks "A B" from a playlist and then delete "B". File cache: "A B"

 You call a second import with tracks "A B C". File cache: "A B C", only "C" is imported.

You call a third import with tracks "A B C D". File cache: "A B C D", "B D" are imported.

This means the "B" deleted before is added back.

spotify_player/src/cli/client.rs
aome510: > I humbly disagree. It is slightly more than what it could be, but its slight increase in complexity reduces extra iterations and allows for important features.
> 
> The logic follows 3 possible states:
> 
> * New import - add all tracks and write to file
> * Old import, no change - check that hash is the same, then do nothing
> * Old import, changes - add new tracks and write to file
> 
> Would there be a way for me to break this down to make it more simple and readable?

Ok, I understand the main idea of the `import` function now. Will give it a try to simplify the codes.

aome510: > Ok, I understand the main idea of the import function now. Will give it a try to simplify the codes.

Created a PR in https://github.com/justjokiing/spotify-player/pull/1.

justjokiing:
Do you think these cli commands warrant a section in the readme? or a link to another md describing these and other commands?

aome510:
> Do you think these cli commands warrant a section in the readme? or a link to another md describing these and other commands?

The commands should be described/explained in the CLI `--help` section. Adding another document page seems to defeat the purpose of the `--help` flag.

spotify_player/src/cli/mod.rs
```diff
@@ -56,6 +56,32 @@ pub enum IdOrName {
...
+    List,
+    Import {
+        from: PlaylistId<'static>,
+        to: PlaylistId<'static>,
+        delete: bool,
+    },
+    Fork {
+        id: PlaylistId<'static>,
+    },
+    Update {
```
aome510, line 79: `update` is confusing. This should be `SyncImport` or `UpdateImport`. When I see `update`, I will expect it to update a playlist's data like name, description, etc.

spotify_player/src/cli/handlers.rs
```diff
@@ -178,8 +179,99 @@ pub fn handle_cli_subcommand(
...
+                .get_one::<String>("id")
+                .expect("Playlist id is required.")
+                .to_owned();
+
+            let id = PlaylistId::from_id(id_s.to_owned())?;
+
+            println!("Forking '{}'...\n", id_s);
+            PlaylistCommand::Fork { id }
+        }
+        "update" => {
```
aome510, line 251: should be `sync_import` or `update_import`

spotify_player/src/cli/handlers.rs
```diff
@@ -178,8 +179,99 @@ pub fn handle_cli_subcommand(
...
+                None
+            };
+
+            let delete = args.get_flag("delete");
+
+            if pid.is_some() {
+                println!("Updating '{}'...\n", id_s.unwrap());
+            } else {
+                println!("Updating all imports...\n");
+            }
```
aome510, line 266: The message should make clear that the command updates/syncs imports for a given playlist/ all playlists.

spotify_player/src/cli/commands.rs
```diff
@@ -128,3 +128,53 @@ pub fn init_like_command() -> Command {
...
+            .arg(Arg::new("from")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new()))
+            .arg(Arg::new("to")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new()))
+            .arg(Arg::new("delete")
+                .short('d')
+                .long("delete")
+                .action(clap::ArgAction::SetTrue)
+                .help("Deletes any tracks that were deleted by imported playlist since last import."))
+            .after_help("Copies all tracks into a playlist. If imported again, it adds new tracks and outputs information on deleted tracks since last import."))
```
aome510, line 166:
```suggestion
                .help("Deletes any previously imported tracks that are no longer in the imported playlist since last import."))
            .after_help("Import data for each playlist is stored inside the application's cache folder. If imported again, the command only adds new tracks since last import."))
```

spotify_player/src/cli/commands.rs
```diff
@@ -128,3 +128,53 @@ pub fn init_like_command() -> Command {
...
+        .subcommand(Command::new("list").about("Lists all user playlists."))
+        .subcommand(Command::new("fork").about("Creates a copy of a playlist and imports it.")
+            .arg(Arg::new("id")
+                .value_parser(clap::builder::NonEmptyStringValueParser::new())))
+        .subcommand(Command::new("update").about("Updates all imported playlists or a single individual import.")
```
aome510, line 171: should be `sync_import` or `update_import`

spotify_player/src/cli/commands.rs
```diff
@@ -128,3 +128,53 @@ pub fn init_like_command() -> Command {
...
+            .arg(Arg::new("id")
+                .required(false)
+                .value_parser(clap::builder::NonEmptyStringValueParser::new()))
+            .arg(Arg::new("delete")
+                .short('d')
+                .long("delete")
+                .action(clap::ArgAction::SetTrue)
+                .help("Deletes any tracks that were deleted by imported playlist since last import.")))
```
aome510, line 179:
```suggestion
                .help("Deletes any previously imported tracks that are no longer in a imported playlist since last import.")))
```

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+            let playlist_check = client
+                .spotify
+                .playlist_check_follow(id.to_owned(), &[uid])
+                .await;
+            if playlist_check.is_err() {
+                anyhow::bail!("Could not find playlist {}", id)
+            }
+
+            // Won't delete if not following
+            let following = playlist_check.unwrap().pop().unwrap();
```
aome510, line 409:
```suggestion
            let following = client
                .spotify
                .playlist_check_follow(id.to_owned(), &[uid])
                .await?.pop().unwrap();
```

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+        } => playlist_import(client, import_from, import_to, delete).await,
+        PlaylistCommand::Fork { id } => {
+            let user = state.data.read().user_data.user.to_owned().unwrap();
+            let uid = user.id;
+
+            let from = client
+                .spotify
+                .playlist(id.to_owned(), None, None)
+                .await
+                .unwrap_or_else(|_| panic!("Cannot import from {}. Playlist not found.", id.id()));
```
aome510, line 445:
```suggestion
            let from = client
                .spotify
                .playlist(id.to_owned(), None, None)
                .await?;
```

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+            let from_desc = from.description.unwrap_or("".to_owned());
```
aome510, line 446:
```suggestion
            let from_desc = from.description.unwrap_or_default();
```

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+        PlaylistCommand::Update { id, delete } => {
+            let user = state.data.read().user_data.user.to_owned().unwrap();
+            let uid = user.id;
+
+            // Get import dir/file
+            let conf_dir = get_cache_folder_path()?;
+            let imports_dir = conf_dir.join("imports");
+
+            // If use specific id option
+            if id.is_some() {
```
aome510, line 478: Don't repeat yourself. Both cases can be combined into a single for loop.

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+            if !track_buff.is_empty() {
+                client
+                    .spotify
+                    .playlist_remove_all_occurrences_of_items(import_to.as_ref(), track_buff, None)
+                    .await?;
+            }
+            result += &format!("Deleted tracks: \n");
+        } else {
+            result += &format!("Import removed tracks: \n");
+        }
```
aome510, line 676: Why do you need to handle this case (no `delete` option is specified)?

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+                    Some(from.collaborative),
+                    Some(from_desc.as_str()),
+                )
+                .await?;
+            let mut result = format!(
+                "Forked {}.\nNew playlist: {}:{}\n",
+                id.id(),
+                to.name,
+                to.id.id()
+            );
```
aome510, line 463:
```suggestion
            let mut result = format!(
                "Forked {}.\nNew playlist: {}:{}\n",
                id.id(),
                to.id.id(),
                to.name,
            );
```

for consistency

aome510 (changes requested):

spotify_player/src/cli/client.rs
```diff
@@ -350,3 +362,359 @@ async fn handle_playback_request(
...
+    let mut new_tracks_hash_set = &from_hash_set - &to_hash_set;
+
+    let mut result = String::new();
+    result += &format!(
+        "Importing tracks from {}:{} to {}:{}...\n",
+        import_from.id(),
+        from_name,
+        import_to.id(),
+        to_name
+    );
```
aome510, line 630: Should be moved after deleting and before importing new tracks.

aome510: An example log: 
```
Importing tracks from 7xASNlY8MoDGeEwfA8yxTH:JP to 19W9UXoblc5JYrKCa85mH0:JP...
Deleting previously imported tracks in JP that are not in the current JP...
Deleted tracks:
    1di1C0QI6Y92yZPYn6XYAZ: シルエット
    7dEfa89dZfo6CQPdsgGCF6: Nandemonaiya - movie ver.
Imported tracks:
    5QiiYSr4pJMPOESQznu2IC: WAYO
    6EY5igURAXBCEen56oKUif: Blue
```
which is confusing when outputs "importing ..." then nothing shows up.

justjokiing: I had moved the "Importing tracks" to the top because it showed what was importing before it gave more information. Having it at the top gives a better look when you are looking at multiple imports.

If it is not at the top, you see deleted tracks, then where the tracks where deleted from, which to me seems backwards. Is there a different message that would convey that shows import is happening before actual details of the import?

spotify_player/src/cli/client.rs
justjokiing: I thought it may be confusing to say "Deleted tracks:" when it they were deleted from the imported playlist and not the import playlist.

spotify_player/src/cli/mod.rs
justjokiing: Is this in reference to the cli command name too?

spotify_player/src/cli/client.rs
aome510: I think we can make it like

```
Importing from "playlist A" to "playlist B"...
Deleting previously imported tracks that are no longer available in "playlist A"...
Deleted tracks:
   ....
Importing new tracks since the last import to "playlist B"...
Imported tracks:
   ....
```

spotify_player/src/cli/client.rs
aome510: I mean you shouldn't handle this case. If no `delete` option is specified, it should not print anything, even information about deleted tracks from the imported playlist.

spotify_player/src/cli/client.rs
aome510: This can fail for many other reasons than playlist not found. The propagated error's message should be enough to identify it, so no need to add additional context.

spotify_player/src/cli/handlers.rs
aome510, line 266:
```suggestion
            if pid.is_some() {
                println!("Updating imports for playlist '{}'...\n", id_s.unwrap());
            } else {
                println!("Updating imports for all playlists...\n");
            }
```

aome510: nit: it's related to what I mentioned before about `update` command.

spotify_player/src/cli/mod.rs
aome510: Yeah, `Sync` works.

aome510:
@justjokiing I saw some comments resolved but nothing changes. Did you push the new changes?

justjokiing:
> @justjokiing I saw some comments resolved but nothing changes. Did you push the new changes?

Yea I planned to commit them with the update command changes but I didn't finish it today. The resolved comments will be in the next commit

justjokiing:
> The commands should be described/explained in the CLI `--help` section. Adding another document page seems to defeat the purpose of the `--help` flag.

I apologize, I more-so meant adding a mention of the Playlist Editing somewhere in the README. It certainly isn't required but I believe that it and other CLI options won't me known to users without it.

justjokiing:
The failed cargo fmt was because of the suggested commit but I will wait to see if there are any other needed changes to commit a fix.

aome510 (approved):
LGTM. Ty for all the back-and-forth and for being patient to address all my comments 🙇 .

I pushed two commits https://github.com/aome510/spotify-player/pull/222/commits/bd6b9408a88381b49bbce35f7f8a49889b2a8006, https://github.com/aome510/spotify-player/pull/222/commits/e4de532e30c11fc8d124215cd0256c8c1ea64796. The former is to update code styles/wordings based on my preferences. The later is to refactor the handling code for `sync` command. Your new implementation did reduce the "DRY"ness but still feel quite verbose to me.

Lmk if those changes don't work for you.

aome510:
> > The commands should be described/explained in the CLI `--help` section. Adding another document page seems to defeat the purpose of the `--help` flag.
> 
> I apologize, I more-so meant adding a mention of the Playlist Editing somewhere in the README. It certainly isn't required but I believe that it and other CLI options won't me known to users without it.

Don't think we need to do this. The command is explorable with `-h` option. For example, running `spotify_player -h`:

```
A command driven spotify player

Usage: spotify_player [OPTIONS] [COMMAND]

Commands:
  get           Get spotify data
  playback      Interact with the playback
  connect       Connect to a Spotify device
  like          Like currently playing track
  authenticate  Authenticate the application
  playlist      Playlist editing
  help          Print this message or the help of the given subcommand(s)

Options:
  -t, --theme <THEME>           Application theme
  -c, --config-folder <FOLDER>  Path to the application's config folder [default: /Users/aome510/.config/spotify-player]
  -C, --cache-folder <FOLDER>   Path to the application's cache folder [default: /Users/aome510/.cache/spotify-player]
  -h, --help                    Print help
  -V, --version                 Print version
```

gives you all commands listed there. You can then run `-h` for each subcommand to explore more.

justjokiing:
> Don't think we need to do this. The CLI is explorable with `-h` option. For example, running `spotify_player -h`:

I think the CLI `-h` option is good and certainly enough. However, this feature is one not seen in other players and would be a selling point for me to use the program. I would not have the chance to know this feature exists without already downloading the program. Though I am obviously biased.

justjokiing:
> LGTM. Ty for all the back-and-forth and for being patient to address all my comments bow .

Thank you for your comments and working with my code. This is my first PR and first large exposure to Rust. Your insight has been incredibly helpful.

aome510:
> > Don't think we need to do this. The CLI is explorable with `-h` option. For example, running `spotify_player -h`:
> 
> I think the CLI `-h` option is good and certainly enough. However, this feature is one not seen in other players and would be a selling point for me to use the program. I would not have the chance to know this feature exists without already downloading the program. Though I am obviously biased.

updated.

justjokiing:
Thank you!

PR #223, opened 2023-07-26
Allow removing a keybind

Kuba314:
This is done by introducing a new `Command::None`, which does nothing and also doesn't appear in the help menu.

To remove a default or any previously defined keybind, simply set its command like this:
```toml
[[keymaps]]
command = "None"
key_sequence = "q"
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -182,7 +182,7 @@ selection = { modifiers = ["Bold", "Reversed"] }
 
 ## Keymaps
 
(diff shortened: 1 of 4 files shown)


aome510 (approved):

Kuba314:
Just fixed the formatting error causing the pipeline to fail.

Kuba314:
Oof, sorry about that. Fixed.

PR #224, opened 2023-07-26
move items in a playlist

### LINKED ISSUE #197: shortcuts for moving items in a playlist?
first of all, epic work! really love it, it only takes 40-55mb ram, i have image & notify enabled, orig. spotify uses up to 800mb, sometimes even more ... crazy

i would wish me a feature, i'm change the order of my songs quite alot (at least at the beginning), would  it be possible to move the songs around?
something like:
"ctrl-arrowUp" & "ctrl-arrowDown" or with shift? or dwm-style with "j" & "k" or whatever makes sense to you ;)

thanks for all the work which went into this project, as soon as i get my pay, i'll sent you a coffee ;)

cheers
hardy

cobbinma:
add track actions to move items up and down in the current playlist 🎵

Screenshot 2023-07-26 at 18 18 30

closes #197

### DIFF
--- a/spotify_player/src/event/window.rs
+++ b/spotify_player/src/event/window.rs
@@ -258,7 +258,6 @@ pub fn handle_command_for_track_list_window(
     mut ui: UIStateGuard,
     state: &SharedState,
 ) -> Result<bool> {
-    tracing::info!("here 0");
     let id = ui.current_page_mut().selected().unwrap_or_default();
     if id >= tracks.len() {
         return Ok(false);
--- a/spotify_player/src/event/window.rs
+++ b/spotify_player/src/event/window.rs
@@ -174,23 +174,8 @@ pub fn handle_command_for_track_table_window(
         }
         Command::ShowActionsOnSelectedItem => {
             let mut actions = command::construct_track_actions(tracks[id], data);
-            if let ContextId::Playlist(playlist_id) = context_id {
+            if let ContextId::Playlist(_) = context_id {
                 actions.push(TrackAction::DeleteFromCurrentPlaylist);
-
-                if let (Some(Context::Playlist { tracks, playlist }), Some(user_id)) = (
-                    state.data.read().caches.context.get(&playlist_id.uri()),
-                    state
-                        .data
-                        .read()
-                        .user_data
-                        .user
-                        .as_ref()
-                        .map(|user| &user.id),
-                ) {
-                    actions.append(&mut track_actions_for_playlist_owner(
-                        tracks, id, user_id, playlist,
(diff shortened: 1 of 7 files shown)


aome510:
Hi @cobbinma, thanks for the PR. I think the original issue wants to implement the feature as commands instead of actions. I can find it quite troublesome, let's say, to move an item up 5 times as you need to call `GetAction`, wait for the popup then select the action. With command, you can just press a key.

It shouldn't be too difficult to change to command.

cobbinma:
> Hi @cobbinma, thanks for the PR. I think the original issue wants to implement the feature as commands instead of actions. I can find it quite troublesome, let's say, to move an item up 5 times as you need to call `GetAction...` command, wait for the popup then select the action. With command, you can just press a key.
> 
> It shouldn't be too difficult to change to command.

Thanks @aome510 

I agree, that is much easier 👍

spotify_player/src/event/window.rs
```diff
@@ -185,6 +185,66 @@ pub fn handle_command_for_track_table_window(
...
+            {
+                if id > 0
+                    && data
+                        .user_data
+                        .playlists
+                        .iter()
+                        .find(|playlist| &playlist.id == playlist_id)
+                        .is_some_and(|playlist| {
+                            Some(&playlist.owner.1)
+                                == data.user_data.user.as_ref().map(|user| &user.id)
```
aome510, line 202: This owner check can be moved inside the `find` function right.

aome510 (changes requested):
Looking good. Only need a minor change

aome510:
I just tried the feature but the ordering does not persist after restart it changes ordering in the UI, which I think mainly because we also update the cached. @cobbinma does it work for you?

cobbinma:
> I just tried the feature but the ordering change does not persist after restart. It does change the ordering in the UI though, which I think mainly because we also update the cached. @cobbinma does it work for you?

@aome510 It's only persisting moving upwards... 😕

aome510:
> It's only persisting moving upwards... 😕

Oh right, also only persists moving upwards for me. Very interesting. Do you think you can investigate this behaviour? Otherwise, I think the PR is in a good state already.

cobbinma:
> > It's only persisting moving upwards... 😕
> 
> Oh right, also only persists moving upwards for me. Very interesting. Do you think you can investigate this behaviour? Otherwise, I think the PR is in a good state already.

@aome510

I misunderstood the difference between insert before and index when using update playlist items.

Should be fixed now 👍

aome510:
Thanks @cobbinma

PR #228, opened 2023-08-01
Add block-image feature which forces rendering images using half blocks

m-torhan:
`spotify-player` with enabled `image` cannot run in `zellij` in `kitty` - it hangs on `viuer::get_kitty_support()`. To enable running `spotify-player` in `zellij` I added feature `block-image` which forces rendering images using half-blocks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,9 @@ cargo install spotify_player --features image
 
 `spotify_player` also supports rendering images with `sixel` behind `sixel` feature flag, which also enables `image` feature:
 
+To disable full resolution images rendering, `spotify_player` needs to be built with `block-image` feature enabled.
+In that case, the images would be rendered using half-blocks.
+This feature enables `image` feature.
 ```shell
 cargo install spotify_player --features sixel
 ```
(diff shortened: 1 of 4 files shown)


aome510:
> `spotify-player` with enabled `image` cannot run in `zellij` in `kitty` - it hangs on `viuer::get_kitty_support()`. To enable running `spotify-player` in `zellij` I added feature `block-image` which forces rendering images using half-blocks.

Does it work in `kitty` outside `zellij` or in `kitty` inside other terminal multiplexers like tmux? It might be `zellij` bug.

That said, I found it weird to add an additional feature to force rendering images as ascii blocks. Is it better to not use `image` feature instead of using one with block rendering?

m-torhan:
@aome510 it works well in `kitty` outside `zellij` (render full-res image). In `tmux` however it renders images using half-blocks, because the `viuer::get_kitty_support()` does not detect `kitty`.

I think having option to render images with ascii blocks when full-res is not available would be nice. I found a workaround for that without any changes in code - simply set `TERM` env var to anything and run `spotify-player`, so if you think this feature is not a good idea, feel free to close this PR.

aome510:
Glad that you find a workaround. IMO, the issue is `zellij`-specific, so it should be fixed from that side. Adding a separate feature on top of `image` to render ascii blocks, which is the "fallback" option if no full-res option is available, feels kinda hacky to me.

I'll close this PR for now. Thanks for raising the issue and taking time to implement this though.

PR #242, opened 2023-08-31
Make it possible to enable streaming only in daemon mode

### LINKED ISSUE #238: Allow creation of streaming device in daemon mode + don't register another device when spawning TUI
Currently if I spawn spotify_player in daemon and then launch another instance in TUI mode - two streaming devices in total are being registered - both with the same name but different IDs

If I set `enable_streaming` to `false` TUI doesn't register new streaming device but then same happens to daemon mode.

I would like to use spotify_player in daemon mode akin to spotifyd and let it register as streaming device.
Then I would like to use spotify_player in TUI mode from time to time in order to control what is being played.

I tend to control spotify device with Raycast plugin and thus I don't need TUI to be opened 100% of the time. 
I use spotify_player in --daemon mode so it registers itself as streaming device.
But from time to time when I want to change playlist I use TUI mode - I would expect TUI to NOT register another streaming device

I wonder if I miss some option to allow such setup or it isn't possible right now.

Schnouki:
Make `enable_streaming` an enum with a `DaemonOnly` option. To keep backward compatibility, `true` and `false` are still accepted too.

Fixes #238.


Schnouki:
(quick disclaimer: I'm extremely new to Rust, this is actually my first attempt at writing Rust code. Sorry if it makes no sense; any feedback would be much appreciated!)

aome510 (approved):
> (quick disclaimer: I'm extremely new to Rust, this is actually my first attempt at writing Rust code. Sorry if it makes no sense; any feedback would be much appreciated!)

Thanks for the PR. LGTM. We all start somewhere. Welcome to Rust 🦀

PR #252, opened 2023-09-13
Set PulseAudio app properties using environment variables

Schnouki:
This allows other PulseAudio clients (like mixers etc.) to show more details about the audio stream.

For example, using the standard PulseAudio volume control (`pavucontrol`):

- before: 

- after:

It would even be possible to do things like add the current track's title and artist in env variables, but I don't think anyone actually makes use of this, so it seems rather pointless to me. Instead, I decided to focus on the most useful ones.

Librespot only sets those when used as a standalone binary, but not as a library: https://github.com/librespot-org/librespot/blob/054074c920d5c6acf0210faa3cd849dc4e065828/src/main.rs#L1211-L1246

Documentation about this PulseAudio feature: https://www.freedesktop.org/wiki/Software/PulseAudio/Documentation/Developer/Clients/ApplicationProperties/


aome510 (approved):
looks good. Thanks

Issue #243, opened 2023-09-02
mute


apprehensions:
**Is your feature request related to a problem? Please describe.**
i don wan to constantly press - and wait a minute cuz volume is slow
 
**Describe the solution you'd like**
a mute keybind

**Describe alternatives you've considered**
muting alsa enitrely

aome510:
> i don wan to constantly press - and wait a minute cuz volume is slow

don't you already have the system mute key? Because of that, it doesn't make sense to implement one for `spotify_player`. `spotify_player` also supports setting volume from CLI: `spotify_player playback volume 0` in case you may need.

apprehensions:
I want to mute specifically spotify-player and not the rest of my system.

PR #253, opened 2023-09-14
Add Mute command

BKasin:
Resolves https://github.com/aome510/spotify-player/issues/243.

Add the ability to immediately set volume to 0% as requested in #243. I bound it to _ since "Shift + -" made sense to me, as it's basically super volume decrease.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ List of supported commands:
 | `Shuffle`                      | toggle the shuffle mode                                                 | `C-s`              |
 | `VolumeUp`                     | increase playback volume by 5%                                          | `+`                |
 | `VolumeDown`                   | decrease playback volume by 5%                                          | `-`                |
+| `Mute`                         | toggle playback volume between 0% and previous level                    | `_`                |
 | `SeekForward`                  | seek forward by 5s                                                      | `>`                |
 | `SeekBackward`                 | seek backward by 5s                                                     | `<`                |
 | `Quit`                         | quit the application                                                    | `C-c`, `q`         |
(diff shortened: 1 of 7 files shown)


aome510:
Hi @BKasin, thanks for the PR. Ideally, `mute` should be a toggle command. Can you update the PR to support that?

BKasin:
@aome510 I agree that that would be preferable, but am not sure how to best implement it.

aome510:
> @aome510 I agree that that would be preferable, but am not sure how to best implement it.

Yeah, it's kinda tricky. One way is to add a new field called `MuteState: Option<u32>` in https://github.com/aome510/spotify-player/blob/master/spotify_player/src/state/player.rs#L5-L15, which is `None` is not mute and `Some(previous_volume_percent)` if is mute. Handling "mute toggle" command will involve updating that state and sending a `PlayerRequest::Volume` request.

BKasin:
Thank you for the guidance, I definitely was over-complicating it mentally.

aome510 (approved):
LGTM! Thanks.

It would be nice to also have UI for the mute functionality. Can you change https://github.com/aome510/spotify-player/blob/05fbda6fe0f5d1e233ef03f8008968f7725e46bf/spotify_player/src/ui/playback.rs#L182-L188 as well?

Maybe  display`volume: old_percent% (muted)` if muted or `volume: current_percent%` if not.

BKasin:
Yeah, a little later today when I get home I'll do that.

aome510:
Thanks @BKasin, after testing the change locally, I thought that the command logic could be simplified a bit more and we also need to handle the case when user increases/decreases the volume on mute. I updated the PR according to implement the above changes in https://github.com/aome510/spotify-player/pull/253/commits/a3930acbca1fb1a77e371b9d11101747e84e407a

BKasin:
Ah, yeah that is definitely a lot better. Thank you!

PR #256, opened 2023-09-20
Consistent Spotify naming

jacksongoode:


spotify_player/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.15.0"
 authors = ["Thang Pham <phamducthang1234@gmail.com>"]
 edition = "2021"
 license = "MIT"
-description = "A command driven spotify player"
+description = "A command driven Spotify player"
```
aome510, line 7: can you update the description to match the Github description as well?

spotify_player/src/main.rs
```diff
@@ -26,7 +26,7 @@ fn init_app_cli_arguments() -> Result<clap::ArgMatches> {
 
     let cmd = clap::Command::new("spotify_player")
         .version("0.15.0")
-        .about("A command driven spotify player")
+        .about("A command driven Spotify player")
```
aome510, line 29: same here

aome510 (changes requested):

aome510:
> I would also change the repo description from:
> 
> ```
> A terminal spotify player that is feature parity with the official client 
> ```
> 
> to
> 
> ```
> A Spotify player in the terminal with full feature parity.
> ```
> 
> as the original phrasing is a bit awkward? Just my suggestion.

Thanks for the suggestion. Updated

spotify_player/Cargo.toml
jacksongoode: Done!

aome510 (approved):

PR #262, opened 2023-09-30
Improve rendering performance for liked tracks page

### LINKED ISSUE #257: Lagginess while scrolling through liked tracks
**Describe the bug**
Scrolling with arrows is laggy on liked tracks list. This doesn't happen on shorter track lists. Wondering what would be causing lagginess though?

**Environment**
- OS: macOS 13
- Application version: Latest
- Application features: default

aome510:
Resolves #257

- enable debug for release profile
- (**BREAKING**) remove `track_table_item_max_len` config option
- improve `is_liked_track` lookup speed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,2 +1,6 @@
 [workspace]
 members = ["spotify_player", "lyric_finder"]
+resolver = "2"
+
+[profile.release]
+debug = 1
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3979,7 +3979,6 @@ dependencies = [
  "tracing",
  "tracing-subscriber",
  "ttl_cache",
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -1,2 +1,6 @@
 [workspace]
 members = ["spotify_player", "lyric_finder"]
+resolver = "2"
+
+[profile.release]
+debug = 1
```
jacksongoode, line 6: @aome510 Would this reduce performance to leave this in for the release build?

aome510: In theory, yes, but I believe the effect should be negligible. Enable debug symbol can help provide better backtrace and make profiling easier.

PR #263, opened 2023-10-02
Add Scoop installation instruction to Readme

rashil2000:
`spotify-player` has been added to Scoop's Main repository: https://github.com/ScoopInstaller/Main/blob/master/bucket/spotify-player.json, making it super easy to install on Windows systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,10 @@ Application's prebuilt binaries can be found in the [Releases Page](https://gith
 
 Run `brew install spotify_player` to install the application.
 
+### Scoop
+
+Run `scoop install spotify-player` to install the application.
+
 ### Cargo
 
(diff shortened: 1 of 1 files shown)


rashil2000:
On a related note, the daemon feature is a unix-only feature, and is not supported on Windows as such. I have modified the Readme to reflect that.

Should I also gate the daemon feature behind a `[target.'cfg(target_os = "unix")'.dependencies]` directive in the Cargo.toml?

aome510:
> Should I also gate the daemon feature behind a `[target.'cfg(target_os = "unix")'.dependencies]` directive in the Cargo.toml?

A readme note saying that the feature is not supported for Windows is fine. I'm not sure how well OS-based dependency and feature-based dependency will work together.

aome510 (approved):

PR #266, opened 2023-10-08
Improve lyrics lookup by removing remix & remaster info in query.

### LINKED ISSUE #250: Genius API is sometimes weird
Getting lyrics from its API behaves a bit weird sometimes.

Searching in their api "Peace Sells Megadeth" show the correct song, but "Peace Sells Remastered Megadeth" does not! And it even does not show Megadeth after the first results.

Don't know how this could be solved.

Maybe we could search for a different API service which we could pass the song and the author as parameters, unlike Genius which is a general term.
Or maybe we could ask the user "is this lyrics wrong" yes -> "please type a search"?

I don't think this has a solution in the very near future, but tracking it could be useful, that's why I created this issue.

By the way, awesome project!! I'm enjoying a lot using it :)

Icelk:
Fixes #250 .

Just simple string manipulation to remove remaster & remix info from query.

I've tried it out and it fixes all my weird lyrics.


lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,45 @@ impl Default for Client {
...
 
+fn improve_query(query: &str) -> String {
+    let is_dash = |c: char| c == '-' || c == '—' || c == '–';
+    let r_no_filler = |s: &str, idx: usize| {
+        let Some(s) = s.get(..idx) else { return idx };
+        s.char_indices()
+            .rev()
+            .find(|(_, c)| !(is_dash(*c) || c.is_whitespace()))
+            .map_or(idx, |(idx, c)| idx + c.len_utf8())
+    };
```
aome510, line 149: what is `r_no_filter` for?

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,45 @@ impl Default for Client {
     }
 }
 
+fn improve_query(query: &str) -> String {
```
aome510, line 141: can you add a function document explaining the function's purpose? Maybe, "Improve the query by filtering out redundant parts, e.g `Remix`, `Remsatered` that may result in incorrect search results"?

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,45 @@ impl Default for Client {
...
+        let Some(s) = s.get(..idx) else { return idx };
+        s.char_indices()
+            .rev()
+            .find(|(_, c)| !(is_dash(*c) || c.is_whitespace()))
+            .map_or(idx, |(idx, c)| idx + c.len_utf8())
+    };
+
+    let mut query = query.to_owned();
+    // just cut xxxx Remaster from query
+    if let Some(remaster_start) = query.find("Remaster").or_else(|| query.find("remaster")) {
```
aome510, line 153: can process the query by `query = query.lower_bound()` so you don't need to manually handle cases like this

aome510 (changes requested):

lyric_finder/src/lib.rs
Icelk: Query is now always converted to lowercase. That shouldn't matter for search results.

lyric_finder/src/lib.rs
Icelk: Now documented in code:
```rust
// reverse finder for non-filler (space, dashes) chars.
// Acts like a trim to remove undesired spaces and dashes.
```

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,60 @@ impl Default for Client {
     }
 }
 
+/// Returns `query` without `remaster` & `remix` information from track/artist query.
+/// Returned value is lowercase.
+/// These caused wildly invalid lyrics to be found.
+/// (try yourself adding remastered 2011 to a song's name when searching in Genius!)
+fn improve_query(query: &str) -> String {
+    let is_dash = |c: char| c == '-' || c == '—' || c == '–';
```
aome510, line 146: this is interesting, doesn't Spotify only use `-` as the dash in most songs? Looking into the codes, it's hard to see the differences between these three cases. IMO, it should be fine to check with `-` only.

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,60 @@ impl Default for Client {
...
+        trim_start = r_no_filler(&query, trim_start);
+        let end = end_of_word(&query, end);
+        query.drain(trim_start..end);
+    }
+    // just cut anything off after - for remix
+    if let Some(remix_start) = query.find("remix") {
+        let end = remix_start + "remix".len();
+        let end = end_of_word(&query, end);
+        if let Some(metadata_start) = query.rfind(is_dash) {
+            if metadata_start > 4 {
```
aome510, line 186: where does the `4` come from?

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,60 @@ impl Default for Client {
...
+    };
+    // used to handle longer variants of words: `remixed`, `remastered`, etc.
+    let end_of_word = |s: &str, idx: usize| {
+        let Some(s) = s.get(idx..) else { return idx };
+        s.find(|c| !c.is_alphanumeric())
+            .map_or(idx, |found| found + idx)
+    };
+
+    let mut query = query.to_lowercase();
+    // just cut xxxx Remaster from query
```
aome510, line 165:
```suggestion
    // remove "xxxx Remaster" from query
    // For example, `{song} xxxx Remastered {artists}` becomes `{song} {artists}`.
```

lyric_finder/src/lib.rs
```diff
@@ -136,6 +138,60 @@ impl Default for Client {
...
+        if let Some(year) = query.get(end_of_prev_word + 1..remaster_start.saturating_sub(1)) {
+            if year.chars().all(|c| c.is_whitespace() || c.is_numeric()) {
+                trim_start = end_of_prev_word;
+            }
+        }
+        trim_start = r_no_filler(&query, trim_start);
+        let end = end_of_word(&query, end);
+        query.drain(trim_start..end);
+    }
+    // just cut anything off after - for remix
```
aome510, line 181:
```suggestion
    // remove "- xxxx remix" from the query
    // For example, `{song} - xxxx remix {artists}` becomes `{song} {artists}`.
```

lyric_finder/src/lib.rs
Icelk: Did it just in case. Will change to your preference :)

lyric_finder/src/lib.rs
Icelk: Documented in the latest commit:
```rust
    // we'll flag for doing something wrong if the new song name is too short.
    const ARBETRARY_SONG_MIN_LENGTH: usize = 3;

```

Icelk:
CI should be working now too.

aome510 (approved):
LGTM! I did a bit of cleanup before merging.

Tbh, after thinking more about the approach, it will be better to handle the removal with regex instead of manually finding the word boundary, and characters or checking year format, which feels kinda hacky.

For the first iteration, I'm good with the approach in this PR.

aome510:
Also, thanks @Icelk for the PR.

Issue #239, opened 2023-08-30
"random" option in CLI Commands


hakuoku:
I already know how to call a random point in a context on the player interface(press `.`), but can you add a simillar feature to CLI Command like [spotify-tui's --random option](https://github.com/Rigellute/spotify-tui#usage) ?
I assume when this feature added the command will be such: `spotify_player playback start context playlist --name "My Playlist" --random`

PR #275, opened 2023-10-22
Allow shuffling context playback from CLI

rudiejd:
Closes https://github.com/aome510/spotify-player/issues/239 

I did this by adding an option that would change the playback context to shuffle. The other option that immediately occurred to me was to shuffle all tracks in the playback locally in the same way that `StartLikedTracks`, but that shifts computation to the client and doesn't achieve the desired result IMO

### DIFF
--- a/spotify_player/src/cli/commands.rs
+++ b/spotify_player/src/cli/commands.rs
@@ -37,6 +37,13 @@ fn init_playback_start_subcommand() -> Command {
                     Arg::new("context_type")
                         .value_parser(EnumValueParser::<ContextType>::new())
                         .required(true),
+                )
+                .arg(
+                    Arg::new("random")
+                        .short('r')
+                        .long("random")
+                        .action(ArgAction::SetTrue)
+                        .help("Shuffle tracks within the launched playback"),
                 ),
         ))
         .subcommand(
--- a/spotify_player/src/cli/commands.rs
+++ b/spotify_player/src/cli/commands.rs
@@ -39,9 +39,9 @@ fn init_playback_start_subcommand() -> Command {
                         .required(true),
                 )
                 .arg(
-                    Arg::new("random")
-                        .short('r')
-                        .long("random")
+                    Arg::new("shuffle")
+                        .short('s')
+                        .long("shuffle")
                         .action(ArgAction::SetTrue)
                         .help("Shuffle tracks within the launched playback"),
                 ),
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -199,10 +199,9 @@ impl Client {
 
                 state.player.write().mute_state = new_mute_state;
             }
-            PlayerRequest::StartPlayback(p, random) => {
-                match random {
-                    Some(random) => playback.shuffle_state = random,
-                    None => (),
+            PlayerRequest::StartPlayback(p, shuffle) => {
+                if let Some(shuffle) = shuffle {
+                    playback.shuffle_state = shuffle;
                 }
                 self.start_playback(p, device_id).await?;
                 // for some reasons, when starting a new playback, the integrated `spotify_player`
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -200,8 +200,10 @@ impl Client {
                 state.player.write().mute_state = new_mute_state;
             }
             PlayerRequest::StartPlayback(p, random) => {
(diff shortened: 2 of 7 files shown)


spotify_player/src/client/mod.rs
```diff
@@ -199,7 +199,9 @@ impl Client {
 
                 state.player.write().mute_state = new_mute_state;
             }
-            PlayerRequest::StartPlayback(p) => {
+            PlayerRequest::StartPlayback(p, random) => {
+                println!("random: {} ", random);
```
aome510, line 203: is this for debugging? If yes, can you remove it?

spotify_player/src/event/window.rs
```diff
@@ -164,12 +164,14 @@ pub fn handle_command_for_track_table_window(
             client_pub.send(ClientRequest::Player(PlayerRequest::StartPlayback(
                 base_playback
                     .uri_offset(tracks[id].id.uri(), state.app_config.tracks_playback_limit),
+                    false
```
aome510, line 167: We should not change the shuffle state of the playback here, either from `false->true` or the opposite, as this behavior is unexpected from the user's perspective.

The `StartPlayback` should take `Option<bool>` instead, and the parameter here is `None`.

spotify_player/src/event/window.rs
```diff
@@ -164,12 +164,14 @@ pub fn handle_command_for_track_table_window(
...
             )))?;
         }
         Command::ChooseSelected => {
             client_pub.send(ClientRequest::Player(PlayerRequest::StartPlayback(
                 base_playback
                     .uri_offset(tracks[id].id.uri(), state.app_config.tracks_playback_limit),
+                    false
```
aome510, line 174: same here

spotify_player/src/event/window.rs
```diff
@@ -265,6 +267,7 @@ pub fn handle_command_for_track_list_window(
             // containing all the tracks in the table.
             client_pub.send(ClientRequest::Player(PlayerRequest::StartPlayback(
                 Playback::URIs(vec![tracks[id].id.clone()], None),
+                false
```
aome510, line 270: same here

spotify_player/src/event/mod.rs
```diff
@@ -27,7 +27,7 @@ pub enum PlayerRequest {
     Volume(u8),
     ToggleMute,
     TransferPlayback(String, bool),
-    StartPlayback(Playback),
+    StartPlayback(Playback, bool),
```
aome510, line 30: should be `Option<bool>`

spotify_player/src/cli/commands.rs
```diff
@@ -37,6 +37,13 @@ fn init_playback_start_subcommand() -> Command {
                     Arg::new("context_type")
                         .value_parser(EnumValueParser::<ContextType>::new())
                         .required(true),
+                )
+                .arg(
+                    Arg::new("random")
```
aome510, line 42: the variable's naming should be `shuffle` instead

aome510 (changes requested):
Thanks for the PR. I would avoid changing the shuffle state unless the user explicitly requests to do so. In that way, `StartPlayback` should set the shuffle state to `true` only when `start playback --shuffle` is called.

aome510:
Maybe an easier way to change is to only set the shuffle state to `true` if `shuffle` is set to true in `StartPlayback` request.

spotify_player/src/event/mod.rs
aome510: Another approach is to set the shuffle state to true if this second parameter is true.

spotify_player/src/client/mod.rs
rudiejd: yep, and removed!

spotify_player/src/client/mod.rs
```diff
@@ -199,7 +199,11 @@ impl Client {
 
                 state.player.write().mute_state = new_mute_state;
             }
-            PlayerRequest::StartPlayback(p) => {
+            PlayerRequest::StartPlayback(p, random) => {
+                match random {
+                    Some(random) => playback.shuffle_state = random,
+                    None => (),
+                }
```
aome510, line 206:
```suggestion
            PlayerRequest::StartPlayback(p, shuffle) => {
                match shuffle {
                    Some(shuffle) => playback.shuffle_state = shuffle,
                    None => (),
                }
```

aome510 (approved):
looks good. One nit

PR #276, opened 2023-10-23
add explicit tags to track information

### LINKED ISSUE #96: Explicit Albums on Artists Page
Depending on the genre of music (like hip-hop), it's commonplace to release 2 versions of an album: an explicit version and an edited "radio" version. When using the Spotify desktop client, you'll see both albums under an Artist page and can pick the one you want. The Explicit songs are marked with an "E" icon.

With `spotify_player`, if I am viewing an Artist's page, I seem to only be shown the "radio" version. I noted I can do a **g**lobal **s**earch with the Album name to find both versions and play the explicit version there.

It would be convenient to access the explicit version from the Artists' Albums page too.

SebRollen:
Fixes #96.

Adds `(Explicit)` to a track name if Spotify reports it as having explicit lyrics. This can make it easier to know if you're listening to the explicit version or the "radio" version of a song.

I originally tried to implement this functionality at the album level, using the `restrictions` field on [the Album API](https://developer.spotify.com/documentation/web-api/reference/get-an-album), but the `explicit` field was not populated in any of my tests. This seems to match the behavior of the spotify web interface, were there is no explicit marker at the album level, but there is one at the track level.

Question for @braheezy (author of #96) and @aome510: what are your thoughts on modifying the track title by appending the string `(Explicit)`? Spotify's web interface obviously handles this a bit differently, adding an `E` symbol, but I wasn't sure how to translate that to the spotify_player interface. Another possiblity could be to add another column in the track-list, populating the column with `E` for explicit tracks and leaving it blank for non-explicit tracks.

## Screenshots:

Explicit album track listing:
Screenshot 2023-10-22 at 8 28 10 PM

"Radio" album track listing:
Screenshot 2023-10-22 at 8 28 24 PM

### DIFF
--- a/spotify_player/src/ui/playback.rs
+++ b/spotify_player/src/ui/playback.rs
@@ -185,7 +185,7 @@ fn render_playback_text(
                         &state.app_config.play_icon
                     },
                     if track.explicit {
-                        format!("{} (Explicit)", track.name)
+                        format!("{} 🅴", track.name)
                     } else {
                         track.name.clone()
                     }
--- a/spotify_player/src/ui/playback.rs
+++ b/spotify_player/src/ui/playback.rs
@@ -185,7 +185,7 @@ fn render_playback_text(
                         &state.app_config.play_icon
                     },
                     if track.explicit {
-                        format!("{} 🅴", track.name)
+                        format!("{} (E)", track.name)
                     } else {
                         track.name.clone()
                     }
--- a/spotify_player/src/ui/playback.rs
+++ b/spotify_player/src/ui/playback.rs
@@ -184,7 +184,11 @@ fn render_playback_text(
                     } else {
                         &state.app_config.play_icon
                     },
-                    track.name,
+                    if track.explicit {
+                        format!("{} (Explicit)", track.name)
+                    } else {
+                        track.name.clone()
+                    }
                 ),
                 ui.theme.playback_track(),
             ),
--- a/spotify_player/src/state/model.rs
+++ b/spotify_player/src/state/model.rs
@@ -245,7 +245,7 @@ impl Track {
     /// gets the track's name, including an explicit label
     pub fn display_name(&self) -> Cow<'_, str> {
         if self.explicit {
-            Cow::Owned(format!("{} 🅴", self.name))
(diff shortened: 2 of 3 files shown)


spotify_player/src/ui/playback.rs
```diff
@@ -184,7 +184,11 @@ fn render_playback_text(
                     } else {
                         &state.app_config.play_icon
                     },
-                    track.name,
+                    if track.explicit {
+                        format!("{} (Explicit)", track.name)
+                    } else {
+                        track.name.clone()
+                    }
```
SebRollen, line 191: The `track` variable here is an `rspotify::FullTrack`, not a `spotify_player::Track`, so I couldn't reuse the display_name method

aome510 (approved):
LGTM! 

> I originally tried to implement this functionality at the album level, using the restrictions field on [the Album API](https://developer.spotify.com/documentation/web-api/reference/get-an-album), but the explicit field was not populated in any of my tests. This seems to match the behavior of the spotify web interface, were there is no explicit marker at the album level, but there is one at the track level.

I actually prefer handling explicitly at the track level because it's possible that a non-explicit album has some explicit tracks.

> what are your thoughts on modifying the track title by appending the string (Explicit)

I go ahead to try using the 🅴 Unicode character instead of (Explicit). Not sure how well it is rendered on other platforms. Tested locally:

Screenshot 2023-10-22 at 9 23 20 PM

SebRollen:
> I go ahead to try using the 🅴 Unicode character instead of (Explicit). Not sure how well it is rendered on other platforms.

Good idea! It looks pretty good on my system too, though a bit hard to read when the track is not highlighted:

Screenshot 2023-10-22 at 10 33 04 PM

I still prefer that approach to the "(Explicit)" text though.

One other symbol for consideration could be the 🇪 unicode character. Looks like this on my system:
Screenshot 2023-10-22 at 10 42 35 PM

aome510:
hmm, look like the unicode is pretty platform-dependent. I think we can either revert back to "(Explicit)" text or "(E)" text for short. I'll merge the PR with "(E)" text.

PR #298, opened 2023-11-25
Allow disabling notify when client is not streaming

VenMolom:
Added `notify_streaming_only` config option. When true, client only sends notification if it has streaming connection.

Coupled with `enable_streaming` set to `DaemonOnly`, this prevents duplicate notifications from being sent when both daemon and UI to control it are running.


aome510 (approved):
looks good, thanks for the PR

PR #305, opened 2023-12-01
UI album year

rubenvandeven:
Thanks for developing spotify-player. I enjoy using it a lot. One thing that was nagging me was that the interface did not give me a sense of time when browsing an artist's page. I therefore added a config option to display the album names prefixed with the year.

As there was no issue mentioning this, I don't know if this implementation fits your ideas for spotify-player, just let me know if it doesn't. Also, my Rust is a bit rusty, so let me know if you have some suggestions before merging this.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -48,6 +48,7 @@ All configuration files should be placed inside the application's configuration 
 | `liked_icon`                         | the icon to indicate the liked state of a song                                           | `♥`                                                       |
 | `border_type`                        | the type of the application's borders                                                    | `Plain`                                                    |
 | `progress_bar_type`                  | the type of the playback progress bar                                                    | `Rectangle`                                                |
+| `album_format`                  | how the name of the album is formatted in the artist's album list                                                    | `YearName`                                                |
 | `playback_window_position`           | the position of the playback window                                                      | `Top`                                                      |
 | `playback_window_width`              | the width of the playback window                                                         | `6`                                                        |
 | `cover_img_width`                    | the width of the cover image (`image` feature only)                                      | `5`                                                        |
@@ -80,6 +81,7 @@ All configuration files should be placed inside the application's configuration 
 - `playback_window_position` can only be either `Top` or `Bottom`.
 - `border_type` can be either `Hidden`, `Plain`, `Rounded`, `Double` or `Thick`.
 - `progress_bar_type` can be either `Rectangle` or `Line`.
+- `album_format` can be either `Name` or `YearName`.
 - `notify_streaming_only=true` and `enable_streaming=DaemonOnly` can be set to avoid sending multiple notifications when both daemon and UI are running.
 
 #### Media control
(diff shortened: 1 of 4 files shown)


aome510 (commented):
Thanks for the PR. I don't think it needs a config option as the release year should be included when displaying albums in the artist page.

spotify_player/src/ui/page.rs
```diff
@@ -618,7 +618,15 @@ fn render_artist_context_page_windows(
...
         let album_items = albums
             .into_iter()
-            .map(|a| (a.name.clone(), false))
+            .map(|a| {
+                (
+                    match state.configs.app_config.album_format {
+                        config::AlbumFormat::Name => a.name.clone(),
+                        config::AlbumFormat::YearName => format!("({1}) {0}", a.name, a.year()),
+                    },
+                    false,
```
aome510, line 627: I prefer `{name} • {year}` for consistency with other display. `name` is placed before `year` as it's the main information.

spotify_player/src/state/model.rs
```diff
@@ -318,6 +318,15 @@ impl Album {
             artists: from_simplified_artists_to_artists(album.artists),
         })
     }
+
+    /// gets the track's album information
```
aome510, line 322:
```suggestion
    /// gets the album's release year
```

docs/config.md
```diff
@@ -48,6 +48,7 @@ All configuration files should be placed inside the application's configuration
 | `liked_icon`                         | the icon to indicate the liked state of a song                                           | `♥`                                                       |
 | `border_type`                        | the type of the application's borders                                                    | `Plain`                                                    |
 | `progress_bar_type`                  | the type of the playback progress bar                                                    | `Rectangle`                                                |
+| `album_format`                  | how the name of the album is formatted in the artist's album list                                                    | `YearName`                                                |
```
aome510, line 51: as I mentioned before, don't need a config option for this. We should always include the year information.

spotify_player/src/ui/page.rs
rubenvandeven: ... or better, create some sort of `render_album_table_window` function similar to  `render_track_table_window` to make the two columns.

aome510: > Thanks @aome510 for the general positive reception. I agree the `•` is best from a consistency point of view.
> 
> Before making the change on the order however, let me share my considerations. I agree name is the main information. However, the length of the album names are vastly different, whereas the year is always only 4 digits. Thus placing the year in front creates a sort of column, which to me looks more clean and allows for much quicker skimming of both name & year. This is convenient because I remember quite some albums by year instead of name. See also the screenshot.
> 
> screenshot_2023-12-05-111757
> 
> Alternatively, if you indeed prefer the year after the name, would you be ok with right aligning the year by left padding it? It would be a bit more elaborate than the current implementation, but reading the code the necessary pad could be calculated from `related_artists_rect.width`.

What you said makes sense. I'm also good with `{year} • {name}` as it looks better. I will try to avoid adding more complexity to implement the year alignment with the `{name} • {year}` format.

rubenvandeven: Perfect! I just removed the config option and left `{year} • {name}` in place. Note that the year is still appended in brackets in the places where the Album `fmt` is used. E.g. in the Albums list of the Library page (see model.rs).

docs/config.md
rubenvandeven: Sure! I removed the option.

aome510 (approved):
LGTM. Thanks. Can you also fix the CI failed?

rubenvandeven:
Thanks for fixing the formatting & merging this change.

PR #329, opened 2024-01-05
Add custom text input widget with cursor, to use in search page

micleo2:
I found using the search bar to be a confusing experience without a cursor. This adds a very simple implementation. You can type, delete with backspace, and navigate with left/right arrows. This can be easily extended in the future to support some basic Emacs-style navigation: by word, end of line etc.

Short video containing usage:
https://github.com/aome510/spotify-player/assets/12721457/5a342c28-d447-4696-ac48-e3d4f0b283bc

### DIFF
--- a/spotify_player/src/event/mod.rs
+++ b/spotify_player/src/event/mod.rs
@@ -2,6 +2,7 @@ use crate::{
     command::{self, Command},
     key::{Key, KeySequence},
     state::*,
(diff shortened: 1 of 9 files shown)


aome510 (approved):
Thank you for the PR. Look great!

Sorry for the late review, I can only have a chance to look into it after my vacation. I also push a small commit to cleanup a few things.

PR #339, opened 2024-01-19
chore: update github actions

LucasFA:
Summary of changes:
- Update several actions to the latest version (actions/checkout, docker/{buildx, login, setup-buildx, build-push-action}
- Migrated actions from actions-rs to dtolnay/rust-toolchain. The former has been archived, the latter has become popular with many larger projects. actions-rs/cargo migrated to run steps.
- Add one rust beta toolchain to CI testing. I only included Ubuntu, figured once is enough. This can be discarded if you want, really. Just to make sure there are no compiler bugs that would soon become an obstable to building the project.
- Add rustc version to actions/cache cache-key, add a restore-keys argument for increased cache hits. Given that target/ is not cached really you can make it hit cache everytime, I think.

I have tested in my own repo the CI and the docker deployment actions. 

IMPORTANT NOTES: 
- the docker/login-action documentation now plain asks, rather than recommends, using a token for access instead of the password. I gather it will still work with a password, though. 
- The CD workflow breaks when updating either the checkout action or the toolchain action. This results in the ubuntu machines cross compiling (and these are the only affected) to break. I've left this as is, but it may be that the containers are not updated anymore; last update was 2 years ago: [link to rustembedded dockerhub page](https://hub.docker.com/r/rustembedded/cross/tags).

Other information: about those warnings in the actions log:
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,10 @@ jobs:
     strategy:
       matrix:
         os: [macOS-latest, windows-latest, ubuntu-latest]
+        toolchain: [stable]
+        include:
+          - os: ubuntu-latest
+            toolchain: beta
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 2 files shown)


aome510 (approved):
Thanks for the PR. I'm not too familiar with GH actions, so really appreciate the improvement!

LucasFA:
Glad to contribute. I will test what can be seamlessly updated in the CD workflow, too.

PR #340, opened 2024-01-21
Fix playback not updating when daemon playback changes

brandon-doornbos:
To reproduce:
1. Run spotify-player as a daemon (`spotify_player -d`)
2. Connect to it with another spotify-player process (`spotify-player`) with `playback_refresh_duration_in_ms` > 0
3. Change the playback of the daemon via the terminal (`spotify_player playback shuffle`)
4. See that the shuffle state only changes in the open process when we manually reload the playback (press `r`) or on a new track

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1305,7 +1305,13 @@ impl Client {
             let new_track = prev_track_name != curr_track_name && !curr_track_name.is_empty();
             // check if we need to update the buffered playback
             let needs_update = match (&player.buffered_playback, &player.playback) {
-                (Some(bp), Some(p)) => bp.device_id != p.device.id || new_track,
+                (Some(bp), Some(p)) => {
+                    bp.device_id != p.device.id
+                        || bp.volume != p.device.volume_percent
+                        || bp.shuffle_state != p.shuffle_state
+                        || bp.repeat_state != p.repeat_state
+                        || new_track
+                }
                 (None, None) => false,
                 _ => true,
             };


spotify_player/src/client/mod.rs
```diff
@@ -1305,7 +1305,13 @@ impl Client {
             let new_track = prev_track_name != curr_track_name && !curr_track_name.is_empty();
             // check if we need to update the buffered playback
             let needs_update = match (&player.buffered_playback, &player.playback) {
-                (Some(bp), Some(p)) => bp.device_id != p.device.id || new_track,
+                (Some(bp), Some(p)) => {
+                    bp.device_id != p.device.id
+                        || bp.volume != p.device.volume_percent
+                        || bp.shuffle_state != p.shuffle_state
+                        || bp.repeat_state != p.repeat_state
+                        || new_track
```
aome510, line 1313: this is not ideal because Spotify takes time to update the playback. The first `shuffle_state`  reported by Spotify API after changing shuffle state will be the old `shuffle_state`. As a result, the **modified** buffered playback will be reset to the old state.

I tried another approach in https://github.com/aome510/spotify-player/pull/341. Can you check if that fixes your issue?

brandon-doornbos: Thank you for working on this. While the approach in #341 does fix the issue when the `playback_refresh_duration_in_ms` > 0, it introduces a flickering of the current playback as can be seen [here](https://asciinema.org/a/tX3vrbY7Kekd8F6HPaM11Mf62).

aome510: update the branch with additional fix for the flickering. You'll need to **force reset** the branch with
```
git fetch && git reset --hard origin
```

brandon-doornbos: This works perfectly, thank you so much for taking the time!

I'll close this PR so you can accept your own.

PR #343, opened 2024-01-23
build: use `cross` for cross-compilation, locally and in CD

LucasFA:
TL;DR: of each commit is in order:
1. Enables local cross compilation from Linux (at least x86_64) to Linux both aarch64 or armv7 by using `cross`.
2. Leverages cross and applies it to the CD workflow to build the release binaries.

Major significant benefit it actually just using current images - the FROM in the Dockerfile resolves to `ghcr.io/cross-rs/aarch64-unknown-linux-gnu:latest`. That by itself could have been a PR.

By the way, to see that TLDR in the code I recommend checking each commit individually. Both together it just looks like a mess 🙃

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -21,49 +21,33 @@ jobs:
             target: aarch64-apple-darwin
           - os: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            container: rust
             dependencies: "libssl-dev libasound2-dev libdbus-1-dev libxcb-shape0-dev libxcb-xfixes0-dev"
           - os: ubuntu-latest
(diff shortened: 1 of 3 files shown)


aome510 (approved):
nice!

LucasFA:
Dropping this here since each bit in the PR is documented within `cross` but is not easily that searchable:
1. [This](https://github.com/cross-rs/cross/wiki/FAQ#github-workflows) is cross' documentation regarding using it in github actions.
2. [This](https://github.com/cross-rs/cross/wiki/Configuration#custom-images) is the custom images section, documenting CROSS_DEB_ARCH and CROSS_DEB_IMAGE

PR #348, opened 2024-01-26
feat: Souvlaki 0.7.0 and Linux Dbus volume control support

LucasFA:
I guess I am just merging other people's work. As I mentioned in [souvlaki's issue for Linux volume control support](https://github.com/Sinono3/souvlaki/issues/36#issuecomment-1911409315), I tested this in my computer with

```
dbus-send --print-reply --dest=org.mpris.MediaPlayer2.spotify_player /org/mpris/MediaPlayer2 org.freedesktop.DBus.Properties.Set string:'org.mpris.MediaPlayer2.Player' string:'Volume' variant:double:0.4
```
(of course change that final 0.4 to different values to appreciate the change in volume)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4106,8 +4106,9 @@ dependencies = [
 
 [[package]]
 name = "souvlaki"
-version = "0.6.0"
-source = "git+https://github.com/Sinono3/souvlaki?branch=volume#0f8e6740b262958408ddc901220c4613b3f517ca"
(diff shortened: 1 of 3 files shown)


aome510 (approved):
thanks for updating this!

LucasFA:
I actually was just going to update the dependency as it didn't break anything here, but then I found this :)

PR #372, opened 2024-02-15
Remove winit from compilation on Linux

LucasFA:
Winit is only used in one line in MacOS and Windows, and unused in Linux.

This change improves from-scratch compile times from ~3m 47s to ~3m 25 on my Linux machine, as `winit` was in the critical chain of compilation. 

I attached the output of `cargo b --release --timings` in case you want to read them. GitHub does not allow uploading HTML files so they're zipped here
[timings.zip](https://github.com/aome510/spotify-player/files/14290010/timings.zip)

(BTW enjoy the vacation!)


aome510 (approved):
LGTM.

> (BTW enjoy the vacation!)

Thank you!

PR #376, opened 2024-02-23
Make Daemon feature imply streaming feature

LucasFA:
Right now launching spotify_player with the `daemon` CLI flag results in a runtime error if the binary was compiled without the streaming feature. As far as I understand it, I do not see why not just do this. The README says it still indicates that it needs _some_ audio backend but AFAIK there's no way to encode that in the feature system.

Note: stumbled into this for the CI suite. There's tooling (aka `cargo hack`) that uses feature dependencies to build many options. This would stop compilation of binaries with the daemon feature without the streaming feature, reducing number of possible checks. Also just general less possible binaries.

### DIFF
--- a/spotify_player/Cargo.toml
+++ b/spotify_player/Cargo.toml
@@ -78,7 +78,7 @@ image = ["viuer", "dep:image"]
 sixel = ["image", "viuer/sixel"]
 notify = ["notify-rust"]
 clipboard = ["copypasta"]
-daemon = ["daemonize"]
+daemon = ["daemonize", "streaming"]
 
 default = ["rodio-backend", "media-control", "clipboard"]
 
--- a/spotify_player/src/main.rs
+++ b/spotify_player/src/main.rs
@@ -283,10 +283,6 @@ fn main() -> Result<()> {
                         eprintln!("Running the application as a daemon on windows/macos with `media-control` feature enabled is not supported!");
                         std::process::exit(1);
                     }
-                    if cfg!(not(feature = "streaming")) {
-                        eprintln!("`streaming` feature must be enabled to run the application as a daemon!");
-                        std::process::exit(1);
-                    }
 
                     tracing::info!("Starting the application as a daemon...");
                     let daemonize = daemonize::Daemonize::new();
--- a/spotify_player/src/streaming.rs
+++ b/spotify_player/src/streaming.rs
@@ -15,6 +15,28 @@ use librespot_playback::{
 use rspotify::model::TrackId;
 use serde::Serialize;
 
+#[cfg(not(any(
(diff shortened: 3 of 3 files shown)


LucasFA:
I suppose that a possible reason for the current state of affairs is to notify the user in the case that they are missing the audio backend (rather than just the streaming feature)

For that we can write a `compile_error! (not(any(rodio_backend, jackaudio_backend, etc)))` instead of the runtime error.

If that sounds good, I'll push it

aome510:
> I suppose that a possible reason for the current state of affairs is to notify the user in the case that they are missing the audio backend (rather than just the streaming feature)
> 
> For that we can write a `compile_error! (not(any(rodio_backend, jackaudio_backend, etc)))` instead of the runtime error.
> 
> If that sounds good, I'll push it

Sounds good. When I try to run the app with `cargo run --no-default-features --features streaming`, which was built successfully, I got a bunch of cryptic error:

```
�1������������b�>�2��������u��=���g���Q���@��0���k��K��.������������������������������m��K�z�$�Y��1�����������������������~��j��X��E��6��,��%��%��)���.���2���6���2��*�%�
�����#��1��>�&�L�2�X�A�f�P�i�\�e�c�]�e�O�k�A�m�1�s�%�~���	��������������������~��r��e�8�`�X�T�t�H��8��'������������������������������-�
�J��a�,�u�0��-��x��e���N��5�����s���d���_���_���_���b���m���}�������������������#��>��T��g�p��{���������������������������(���?���Q���`���o���x�����⏎
```

Improve the error message (with compiler error) when user runs `streaming` w/o audio backend is a nice addition.

LucasFA:
Just force pushed a commit for that. For reference, `cargo check --no-default-features --features streaming` now returns:
```
error: Streaming feature is enabled but no audio backend has been selected. Consider adding one of the following features:
           rodio-backend,
           alsa-backend,
           pulseaudio-backend,
           portaudio-backend,
           jackaudio-backend,
           rodiojack-backend,
           sdl-backend,
           gstreamer-backend
       For more information, visit https://github.com/aome510/spotify-player?tab=readme-ov-file#streaming
  --> spotify_player/src/streaming.rs:28:1
   |
28 | / compile_error!("Streaming feature is enabled but no audio backend has been selected. Consider adding one of the following features:
29 | |     rodio-backend,
30 | |     alsa-backend,
31 | |     pulseaudio-backend,
...  |
37 | | For more information, visit https://github.com/aome510/spotify-player?tab=readme-ov-file#streaming
38 | | ");
   | |__^

error: could not compile `spotify_player` (bin "spotify_player") due to 1 previous error
```

aome510 (approved):

PR #385, opened 2024-03-02
Extended CI suite

LucasFA:
Took me a while to get around to it, but finally there's something to show for it.

- Set a MSRV in the Cargo.toml and a new MSRV job
- Check OS specific features, both in MSRV and general CI job
- Cache `./target/` and creates a CI profile to reduce its cache usage, though Swatinem does a good job already 
- Reorders calls to `cargo clippy` and `test` to have the quickest at the beginning

I also took the liberty of reordering the features in the Cargo.toml file, in decreasing number of dependencies, mostly. We can drop that.
New (action) dependencies: [taiki-e/install-action@v2](https://github.com/taiki-e/install-action) and [Swatinem/rust-cache](https://github.com/Swatinem/rust-cache)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ env:
   CARGO_TERM_COLOR: always
   RUST_BACKTRACE: 1
   RUST_FEATURES: "rodio-backend,lyric-finder,media-control,image,notify,clipboard"
+  LINUX_PACKAGES: "libssl-dev libasound2-dev libdbus-1-dev libxcb-shape0-dev libxcb-xfixes0-dev"
 
 jobs:
    msrv:
@@ -26,7 +27,7 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-      - run: sudo apt-get update && sudo apt-get install libssl-dev libasound2-dev libdbus-1-dev libxcb-shape0-dev libxcb-xfixes0-dev
+      - run: sudo apt-get update && sudo apt-get install ${{ env.LINUX_PACKAGES }}
         if: ${{ runner.os == 'Linux'}}
       - uses: actions/checkout@v4
       
@@ -58,7 +59,7 @@ jobs:
 
     steps:
       - name: Install dependencies
-        run: sudo apt-get update && sudo apt-get install libssl-dev libasound2-dev libdbus-1-dev libxcb-shape0-dev libxcb-xfixes0-dev
+        run: sudo apt-get update && sudo apt-get install ${{ env.LINUX_PACKAGES }}
         if: ${{ runner.os == 'Linux' }}
 
       - name: Checkout source
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,33 @@ env:
   RUST_FEATURES: "rodio-backend,lyric-finder,media-control,image,notify,clipboard"
 
 jobs:
-  rust-ci:
+   msrv:
+    if: github.event.pull_request.draft != true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest, windows-latest, ubuntu-latest]
+        include:
+          - os: ubuntu-latest
+            OS_SPECIFIC_FEATURES: ",daemon, sixel"
+          - os: macOS-latest
+            OS_SPECIFIC_FEATURES: ",sixel"
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - run: sudo apt-get update && sudo apt-get install libssl-dev libasound2-dev libdbus-1-dev libxcb-shape0-dev libxcb-xfixes0-dev
+        if: ${{ runner.os == 'Linux'}}
+      - uses: actions/checkout@v4
+      
+      - uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hack
+
+      - uses: Swatinem/rust-cache@v2
+
+      - run: cargo hack check --rust-version --package spotify_player --all-targets --profile ci --features "${{ env.RUST_FEATURES }} ${{ matrix.OS_SPECIFIC_FEATURES }}"
+
+   rust-ci:
     if: github.event.pull_request.draft != true
 
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,11 +48,11 @@ jobs:
         include:
           - os: ubuntu-latest
             toolchain: beta
-            OS_SPECIFIC_FEATURES: ",daemon, sixel"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -12,7 +12,33 @@ env:
   RUST_FEATURES: "rodio-backend,lyric-finder,media-control,image,notify,clipboard"
```
aome510, line 12: I realize once you surround the arguments with quotes, spaces can be used instead of commas. See my below suggestions

.github/workflows/ci.yml
```diff
@@ -12,7 +12,33 @@ env:
...
+    if: github.event.pull_request.draft != true
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macOS-latest, windows-latest, ubuntu-latest]
+        include:
+          - os: ubuntu-latest
+            OS_SPECIFIC_FEATURES: ",daemon, sixel"
+          - os: macOS-latest
+            OS_SPECIFIC_FEATURES: ",sixel"
```
aome510, line 25:
```suggestion
          - os: ubuntu-latest
            OS_SPECIFIC_FEATURES: "daemon sixel"
          - os: macOS-latest
            OS_SPECIFIC_FEATURES: "sixel"
```

.github/workflows/ci.yml
```diff
@@ -12,7 +12,33 @@ env:
   RUST_FEATURES: "rodio-backend,lyric-finder,media-control,image,notify,clipboard"
 
 jobs:
-  rust-ci:
+   msrv:
+    if: github.event.pull_request.draft != true
+    strategy:
+      fail-fast: false
```
aome510, line 18: why not `fail-fast` on this one?

.github/workflows/ci.yml
```diff
@@ -12,7 +12,33 @@ env:
   RUST_FEATURES: "rodio-backend,lyric-finder,media-control,image,notify,clipboard"
 
 jobs:
-  rust-ci:
+   msrv:
```
aome510, line 15: can `msrv` be put as a step of `rust-ci` instead to avoid code duplication between those two?

.github/workflows/ci.yml
```diff
@@ -22,6 +48,11 @@ jobs:
         include:
           - os: ubuntu-latest
             toolchain: beta
+            OS_SPECIFIC_FEATURES: ",daemon, sixel"
+          - os: ubuntu-latest
+            OS_SPECIFIC_FEATURES: ",daemon, sixel"
+          - os: macOS-latest
+            OS_SPECIFIC_FEATURES: ",sixel"
```
aome510, line 55:
```suggestion
            OS_SPECIFIC_FEATURES: "daemon sixel"
          - os: ubuntu-latest
            OS_SPECIFIC_FEATURES: "daemon sixel"
          - os: macOS-latest
            OS_SPECIFIC_FEATURES: "sixel"
```

.github/workflows/ci.yml
aome510: then you also need to update this variable to use spaces

Cargo.toml
```diff
@@ -2,5 +2,14 @@
 members = ["spotify_player", "lyric_finder"]
 resolver = "2"
 
+[workspace.package]
+rust-version = "1.74"
```
aome510, line 6: will it return a compile error if you have an older version than this one?

Cargo.toml
```diff
@@ -2,5 +2,14 @@
...
+
+[profile.ci]
```
aome510, line 8: what is `ci` profile for? How is it different from `dev`?

aome510 (commented):
Have a few questions. Do you also know why the last job's name is weird?

image

Cargo.toml
LucasFA: Setting `debug = 0` avoids emitting a large amount of debug information which is largely not needed when just checking if the crate can compile. That way the `./target/` cache is smaller. Likewise with `incremental` and `strip`. I was mostly following [matklad's recommendations on the topic](https://matklad.github.io/2021/09/04/fast-rust-builds.html#CI-Workflow). 

I admittedly am not sure about `incremental`, as it does reduce the size but I do not know about its performance implications, which in any case would be limited to compiling the spotify_player and lyric_finder crates as the dependencies are all cached.

For a ballpark estimation, I just run `cargo check` (debug profile) and `cargo check --profile ci` in my Linux machine to compare the sizes of the output: the former was 665M, the latter 367M.

Cargo.toml
LucasFA: As it is, yes, I get the following:
```
error: package `clap_lex v0.7.0` cannot be built because it requires rustc 1.74 or newer, while the currently active rustc version is 1.73.0
Either upgrade to rustc 1.74 or newer, or use
cargo update -p clap_lex@0.7.0 --precise ver
where `ver` is the latest version of `clap_lex` supporting rustc 1.73.0
```

I don't know if there are dependencies unrelated to clap that also ask for >=1.74

.github/workflows/ci.yml
LucasFA: I think that being a `check` rather than `build` or `test`, the job should be fast enough to not matter for a few seconds, and adds the feedback on whether it broke for every platform or not - likely would be a platform dependant dependency. 
Though if keeping the MSRV for builds with all (compatible) features it would have to be bumped anyway.

.github/workflows/ci.yml
LucasFA: I'm seeing if the solutions [in this discussion](https://github.com/orgs/community/discussions/26284) to reuse matrices can be applied in a simple way here. Notable possible problem is that they don't use the same matrices as it is right now.
That's your main concern I assume, though the checkout and install packages steps could be abstracted away into another workflow, too.

Cargo.toml
aome510: got it, make sense to me. Thanks for the context

.github/workflows/ci.yml
aome510: I thought you only need to move 
```
- name: Install cargo-hack
  uses: taiki-e/install-action@cargo-hack

- name: Run MSRV check 
  run: cargo hack check --rust-version --package spotify_player --all-targets --profile ci --features "${{ env.RUST_FEATURES }} ${{ matrix.OS_SPECIFIC_FEATURES }}"
```
to be in `rust-ci.steps`.

aome510: yeah, one needs to check the action's logs if CI run fails anyway.

LucasFA:
Alright I'm making this a draft - I'm trying to figure a way without repetitions but that keeps the jobs separate. I'm working on another (personal) branch - be back in a few days

PR #432, opened 2024-04-25
add CLI search command

### LINKED ISSUE #425: Search from CLI
No

Right now the current method for getting search-like information from Spotify via the CLI is to use `spotify_player get item <TYPE> --name`. Unfortunately this only returns a single result rather than the list of results that the TUI returns when searching. 

A new CLI command called `search` which allows users to do the same kind of search that the TUI does.

N/A

**Additional context**
From poking through the codebase (I'm not super familiar with Rust so I'm sure I've missed something here) it looks like the CLI does item retrieval calls out directly to the `artist`, `track`, `playlist`, and `track` methods in rspotify.  Would it be too much work to wire up the `search` method as well?

kadinsayani:
Resolves #425

### DIFF
--- a/spotify_player/src/cli/client.rs
+++ b/spotify_player/src/cli/client.rs
@@ -165,7 +165,7 @@ async fn handle_socket_request(
             let resp = handle_playlist_request(client, command).await?;
             Ok(resp.into_bytes())
         }
-        Request::Search(query) => {
+        Request::Search { query } => {
             let resp = handle_search_request(client, query).await?;
             Ok(resp)
         }
--- a/spotify_player/src/cli/client.rs
+++ b/spotify_player/src/cli/client.rs
@@ -310,6 +310,39 @@ async fn handle_get_item_request(
     })
 }
 
+async fn handle_search_request(client: &Client, query: String) -> Result<Vec<u8>> {
+    let search_result = client.search(&query).await?;
+
+    let mut results = vec![];
+
+    if let SearchResult::Tracks(tracks) = search_result.tracks {
+        let tracks_json = serde_json::to_string(&tracks)?;
+        results.push(tracks_json.into_bytes());
+    }
+
+    if let SearchResult::Artists(artists) = search_result.artists {
+        let artists_json = serde_json::to_string(&artists)?;
+        results.push(artists_json.into_bytes());
+    }
+
+    if let SearchResult::Albums(albums) = search_result.albums {
+        let albums_json = serde_json::to_string(&albums)?;
+        results.push(albums_json.into_bytes());
+    }
+
+    if let SearchResult::Playlists(playlists) = search_result.playlists {
+        let playlists_json = serde_json::to_string(&playlists)?;
+        results.push(playlists_json.into_bytes());
+    }
+
+    let mut combined_results = vec![];
+    for result in results {
+        combined_results.extend(result);
+    }
+
+    Ok(combined_results)
+}
+
 async fn handle_playback_request(
     client: &Client,
     state: &Option<SharedState>,
--- a/spotify_player/src/cli/client.rs
+++ b/spotify_player/src/cli/client.rs
@@ -165,6 +165,10 @@ async fn handle_socket_request(
             let resp = handle_playlist_request(client, command).await?;
             Ok(resp.into_bytes())
         }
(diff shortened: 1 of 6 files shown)


spotify_player/src/cli/client.rs
```diff
@@ -310,6 +310,39 @@ async fn handle_get_item_request(
...
+        let playlists_json = serde_json::to_string(&playlists)?;
+        results.push(playlists_json.into_bytes());
+    }
+
+    let mut combined_results = vec![];
+    for result in results {
+        combined_results.extend(result);
+    }
+
+    Ok(combined_results)
```
aome510, line 343: I don't think the above codes will compile as `Client::search` returns a struct `SearchResults` as follows:

```rust
#[derive(Default, Clone, Debug)]
/// Data returned when searching a query using Spotify APIs.
pub struct SearchResults {
    pub tracks: Vec<Track>,
    pub artists: Vec<Artist>,
    pub albums: Vec<Album>,
    pub playlists: Vec<Playlist>,
}
```

you can add `#derive(Deserialize, Serialize)` for `SearchResults` so that it can be serialized to json bytes directly using `serde_json::to_vec`. Then the `handle_search_request` will be really simple

```suggestion
    let search_result = client.search(&query).await?;
    Ok(serde_json::to_vec(&search_result)?)
```


Maybe you don't need to define a separate function to handle search.

spotify_player/src/cli/handlers.rs
```diff
@@ -210,6 +210,7 @@ pub fn handle_cli_subcommand(cmd: &str, args: &ArgMatches) -> Result<()> {
         "like" => Request::Like {
             unlike: args.get_flag("unlike"),
         },
+        // TODO: handle search
```
aome510, line 213: the pattern to handle a CLI command is 
1. defining a command like you already did in `spotify_player/src/cli/commands.rs`
2. defining a request variant of `Request`  in `spotify_player/src/cli/mod.rs`, for example `Search { query: string }`
3. construct a request based on CLI arguments, see how `Request::Like` is constructed above. For example, `Request::Search { query: args.get_one::<String>("query").expect("query is required").to_owned() }`
4. handle a request in `cli/client.rs`, you already defined a `handle_search_request` function

aome510: you can take a look into how the app handles `get` command or other commands to understand the pattern

spotify_player/src/cli/client.rs
kadinsayani: Thanks for the suggestion!

spotify_player/src/cli/handlers.rs
kadinsayani: Got it. Thanks for your suggestions and help on this :)

spotify_player/src/cli/client.rs
```diff
@@ -165,6 +165,10 @@ async fn handle_socket_request(
             let resp = handle_playlist_request(client, command).await?;
             Ok(resp.into_bytes())
         }
+        Request::Search(query) => {
```
aome510, line 168:
```suggestion
        Request::Search { query } => {
```

spotify_player/src/cli/commands.rs
```diff
@@ -123,6 +123,16 @@ pub fn init_playback_subcommand() -> Command {
         )
 }
 
+pub fn init_search_command() -> Command {
+    Command::new("search").about("Search spotify").arg(
+        Arg::new("query")
+            .long("query")
+            .short('q')
```
aome510, line 130:
```suggestion
```

doesn't need to be an option, e.g it's better to call `search "query"` than `search --query "query"`

aome510 (commented):
looking good

spotify_player/src/cli/commands.rs
```diff
@@ -123,6 +123,15 @@ pub fn init_playback_subcommand() -> Command {
         )
 }
 
+pub fn init_search_command() -> Command {
+    Command::new("search").about("Search spotify").arg(
+        Arg::new("query")
+            .help("Search query")
+            .required(true)
+            .index(1),
```
aome510, line 131: I don't think `index` is needed here

aome510 (approved):

PR #441, opened 2024-04-29
New AUR package - adding to README?

blackthorne7:
hi! i created an [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=spotify-player-full-pipe) that compiles with all the disabled features exposed. It also compiles with Pulseaudio/Pipewire as the audio backend instead of rodio.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Run `cargo install spotify_player` to install the application from [crates.io](h
 
 Run `yay -S spotify-player` to install the application as an AUR package.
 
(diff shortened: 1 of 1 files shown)


aome510:
Thanks for the PR.

>  is there a reason why these features are disabled by default?

More features mean more stuff to compile and some don't work well in other systems, e.g. `image` doesn't render well on some terminals, `media-control` requires spawning a separate window, etc.

PR #466, opened 2024-06-13
Add Alternating Colors for Help Table

przb:
I started using spotify-player, and I found the help table difficult to read. I added alternating row coloring as pictured below to help with identifying keybindings. 

image

### DIFF
--- a/examples/theme.toml
+++ b/examples/theme.toml
@@ -43,6 +43,31 @@ bright_cyan = "#8be9fd"
 bright_white = "#ffffff"
 
 [[themes]]
+name = "my_dracula"
+[themes.palette]
+background = "#1e1f29"
+foreground = "#f8f8f2"
+black = "#000000"
+red = "#ff5555"
+green = "#50fa7b"
+yellow = "#f1fa8c"
+blue = "#bd93f9"
+magenta = "#ff79c6"
+cyan = "#8be9fd"
+white = "#bbbbbb"
+bright_black = "#555555"
+bright_red = "#ff5555"
+bright_green = "#50fa7b"
+bright_yellow = "#f1fa8c"
+bright_blue = "#bd93f9"
+bright_magenta = "#ff79c6"
+bright_cyan = "#8be9fd"
+bright_white = "#ffffff"
+[themes.component_style]
+selection = {bg = "Black", fg = "White", modifiers = ["Bold"]}
+secondary_row = { bg = "#677075" }
+
+[[themes]]
 name = "gruvbox_dark"
 [themes.palette]
 background = "#282828"
--- a/examples/theme.toml
+++ b/examples/theme.toml
@@ -42,8 +42,9 @@ bright_magenta = "#ff79c6"
 bright_cyan = "#8be9fd"
 bright_white = "#ffffff"
 
+# same colors as dracula, but includes non-default component styles
 [[themes]]
-name = "my_dracula"
+name = "dracula2"
 [themes.palette]
 background = "#1e1f29"
 foreground = "#f8f8f2"
@@ -64,7 +65,7 @@ bright_magenta = "#ff79c6"
 bright_cyan = "#8be9fd"
 bright_white = "#ffffff"
 [themes.component_style]
-selection = {bg = "Black", fg = "White", modifiers = ["Bold"]}
+selection = { bg = "Black", fg = "White", modifiers = ["Bold"] }
 secondary_row = { bg = "#677075" }
 
 [[themes]]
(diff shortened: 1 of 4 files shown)


aome510 (commented):
Thanks for the PR. Can you make this configurable for both the alternative behaviour and the color?

spotify_player/src/ui/page.rs
```diff
@@ -567,12 +567,19 @@ pub fn render_commands_help_page(frame: &mut Frame, ui: &mut UIStateGuard, rect:
...
+            .enumerate()
+            .map(|(i, (command, keys))| {
                 Row::new(vec![
                     Cell::from(format!("{command:?}")),
                     Cell::from(format!("[{keys}]")),
                     Cell::from(command.desc()),
                 ])
+                    // adding alternating row colors
+                    .style(if (i + scroll_offset) % 2 == 0 {
+                        ui.theme.app().bg(Color::DarkGray)
```
aome510, line 579: should be a component style: https://github.com/aome510/spotify-player/blob/master/docs/config.md#component-styles

przb: Did you have a particular style in mind? Almost seems to me like having a new alternate row field for the ComponentStyle struct could work.

aome510: yeah, I think you could add new component style named `help_table_alternating_row` for example. It's to allow people with incompatible theme to update the style.

spotify_player/src/config/theme.rs
```diff
@@ -74,6 +77,7 @@ pub struct ComponentStyle {
     pub playlist_desc: Option<Style>,
     pub table_header: Option<Style>,
     pub selection: Option<Style>,
+    pub secondary_row: Option<Style>,
```
przb, line 80: I named this secondary row because I thought it could be used in more places than just the help table, but let me know if it should just be for the help table.

przb (commented):
The [docs](https://github.com/aome510/spotify-player/blob/master/docs/config.md) have not been updated yet. Is that manually updated, or should I update it?

spotify_player/src/config/theme.rs
```diff
@@ -58,6 +58,9 @@ pub struct Palette {
     pub bright_blue: Color,
     #[serde(default = "Color::bright_yellow")]
     pub bright_yellow: Color,
+
+    #[serde(default = "Color::dark_gray")]
+    pub dark_gray: Color,
```
przb, line 63: I added support for dark gray coloring, but is there a better way to add support for coloring modifications? Dark and Bright colors seem to have some repetition. And should other dark colors be added as well?

aome510: No you shouldn't add new field to the palette, it represents the basic 16-color palette with additional fields for background and foreground.

spotify_player/src/config/theme.rs
```diff
@@ -284,6 +289,14 @@ impl Theme {
             Some(s) => s.style(&self.palette),
         }
     }
+    pub fn secondary_row(&self) -> tui::style::Style {
+        match &self.component_style.secondary_row {
+            None => Style::default()
+                .bg(StyleColor::DarkGray)
```
aome510, line 295: can use `StyleColor::Rgb` instead if you want a specific color that is not available in the palette? My suggestion would be using default theme if `None`, so it doesn't cause any surprise for current users.

przb: Correct me if I'm wrong, but this enables the alternating row colors by default, right? I was thinking that should be the intended behavior; should it instead only set alternating row colors when using a custom theme?

aome510: > I was thinking that should be the intended behavior; should it instead only set alternating row colors when using a custom theme?

Yeah, it should be opt-in. My reasoning is that setting to some default style (e.g dark gray as above) might "break" some users' theme. If one wants the alternating colors, they can update the style in their configs.

examples/theme.toml
```diff
@@ -42,6 +42,31 @@ bright_magenta = "#ff79c6"
 bright_cyan = "#8be9fd"
 bright_white = "#ffffff"
 
+[[themes]]
+name = "my_dracula"
```
aome510, line 46: nit: can you name it `dracula2` or something? Can also add a small comment explaining the difference between this and the previous `dracula`

apprehensions:
Why not introduce this to the music page as well?

aome510:
> Why not introduce this to the music page as well?

Do you mean track table? Should be easy to add tho

PR #474, opened 2024-06-25
fix #464 by adding full date to display next to albums

### LINKED ISSUE #464: Artist's albums of the same year disappear
**Describe the bug**
In the Album pane of the artist page albums that have the same year (and possibly same year) show only once. 

**To Reproduce**
For example, I encountered this looking for Audioslave's Revelations album. It seems to show up on the artist page, but selecting that album refers to an album dated 2006-09-01 with just a single track. I suppose this an EP.

Searching for Revelations does yield the full album and I can open it from there. It is of the same year, but a different day: 2006-08-29. See also the images below.

**Expected behaviour**
Both albums should show. Ideally they have an indicator how they're different, but as a first step simply having both would already be great.

**Screenshots**

Overview with just one album:
screenshot_2024-06-12-172612

The selected album just has one track. Album date: 2006-09-01
screenshot_2024-06-12-172620

Album with the same artist, name & year, but different month and day: 2006-08-29. This can be accessed through search.
screenshot_2024-06-12-172557

**Environment**

- OS: Linux - Debian testing (Trixie)
- Application version:  0.18.2 (installed through cargo)
- Application features: `cargo install spotify_player --no-default-features --features pulseaudio-backend,streaming,notify,media-control,sixel`
screenshot_2024-06-12-172620
screenshot_2024-06-12-172612
screenshot_2024-06-12-172557

DOD-101:
fixes #464

This PR removes the code previously responsible for filtering out duplicate album names, which had caused #464.

I have decided to not any additional text to signal the difference between the albums with the same name as: 
1) That might cause confusion as the the actual name of the album
2) There is in my opinion no problem with having 2 entries with the same name
3) The official Spotify client also handles this situation in this manner

This PR however doesn't entirely fix the bug described in the issue, as it seems that this is an upstream issue. The Spotify API doesn't return both copies of the album "Revelations" by "Audioslave" and accordingly the Spotify Client has the exact same behavior as spotfiy_player. But now the two versions of Original Fire show up as they should. 

[Spotify response sample](https://developer.spotify.com/documentation/web-api/reference/get-an-artists-albums): 

`jq '.items[].name' response.json`

```
"Revelations"
"Out of Exile"
"Audioslave"
"Original Fire"
"Original Fire"
"Doesn't Remind Me"
"Sessions @AOL Music - EP (Live)"
"Your Time Has Come"
"Like A Stone (Live from New York City)"
"Be Yourself"
"Audioslave Live on Hollywood Blvd."
"Show Me How To Live"
"Cochise"
"Chris Cornell (Deluxe Edition)"
"Live 8 (Live, July 2005)"
"Rock Clássico"
"90s Grunge"
"Canciones De Hace Un Tiempo - Rock"
"20 Top Rock Hits"
"Summer Rocks Best Buy Version"
"Miami Ibiza St. Tropez, Selected Housetunes Vol. 1"
```

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1538,21 +1538,13 @@ impl Client {
 
     /// Process a list of albums, which includes
     /// - sort albums by the release date
-    /// - remove albums with duplicated names
     fn process_artist_albums(&self, albums: Vec<Album>) -> Vec<Album> {
         let mut albums = albums.into_iter().collect::<Vec<_>>();
 
         albums.sort_by(|x, y| x.release_date.partial_cmp(&y.release_date).unwrap());
 
-        // use a HashSet to keep track albums with the same name
-        let mut seen_names = std::collections::HashSet::new();
+        albums.reverse();
 
-        albums.into_iter().rfold(vec![], |mut acc, a| {
-            if !seen_names.contains(&a.name) {
-                seen_names.insert(a.name.clone());
-                acc.push(a);
-            }
-            acc
-        })
+        albums
     }
 }
--- a/spotify_player/src/ui/page.rs
+++ b/spotify_player/src/ui/page.rs
@@ -728,7 +728,7 @@ fn render_artist_context_page_windows(
     let (album_list, n_albums) = {
         let album_items = albums
             .into_iter()
-            .map(|a| (format!("{1} • {0}", a.name, a.year()), false))
+            .map(|a| (format!("{1} • {0}", a.name, a.release_date), false))
             .collect::<Vec<_>>();
 
         utils::construct_list_widget(


aome510:
> I have decided to not any additional text to signal the difference between the albums with the same name as:
> 
> That might cause confusion as the the actual name of the album
> There is in my opinion no problem with having 2 entries with the same name
> The official Spotify client also handles this situation in this manner

Maybe you can add [`album_type`](https://docs.rs/rspotify-model/0.13.2/rspotify_model/enums/types/enum.AlbumType.html) and render it in UI to better differentiate between albums of the same name. Official Spotify app does this.

Another approach would be updating the UI to render the full date instead of just year

DOD-101:
Here are the two options you suggested:

 1)
image

This one doesn't fix the issue of ambiguity you addressed.

 2)
image

This is, in my opinion, too much detail / text if you are just looking for an album.

\+ two other options

3)
image

This doesn't directly show the exact distinction between the two albums, but does at least show that there is some kind of difference. 

4) 
Another possible solution would be to display the number of songs in an album. This is probably the most preferable solution, as it would in most cases signal a clear difference between the different versions without adding too much info or text that might not be useful to the end user just looking for an album. 

I'll work on implementing option 4, unless you raise any objections with it.

aome510:
> Here are the two options you suggested:
> 
> 
> image
> 
> This one doesn't fix the issue of ambiguity you addressed.
> 
> 
> image
> 
> This is, in my opinion, too much detail / text if you are just looking for an album.
> 
> + two other options
> 
> 
> image
> 
> This doesn't directly show the exact distinction between the two albums, but does at least show that there is some kind of difference.
> 
> 
> Another possible solution would be to display the number of songs in an album. This is probably the most preferable solution, as it would in most cases signal a clear difference between the different versions without adding too much info or text that might not be useful to the end user just looking for an album.
> 
> The only problem I have with 4 is that [SimplifiedAlbum](https://docs.rs/rspotify-model/0.13.2/rspotify_model/album/struct.SimplifiedAlbum.html) doesn't contain [tracks](https://docs.rs/rspotify-model/0.13.2/rspotify_model/album/struct.FullAlbum.html#structfield.tracks).

Thanks for the investigation. Really helpful. I would probably go with 2 as it might also help with filtering in case you want to find specific album.

aome510 (approved):

PR #475, opened 2024-06-26
Configurable seek duration

jorismertz:


spotify_player/src/config/mod.rs
```diff
@@ -108,6 +108,8 @@ pub struct AppConfig {
 
     #[cfg(all(feature = "streaming", feature = "notify"))]
     pub notify_streaming_only: bool,
+
+    pub seek_duration_secs: u16,
```
aome510, line 112: can you update https://github.com/aome510/spotify-player/blob/master/examples/app.toml as well?

jorismertz: right ofcourse

aome510 (approved):

Issue #278, opened 2023-10-26
Adjustable column size


cyberfeign:
**Is your feature request related to a problem? Please describe.**
I use this awesome client in a Tmux pane on a custom panel with other applications where space is rather scarce. I've noticed that for instance track duration is sometimes clipped when dimensions are reduced.

**Describe the solution you'd like**
I'd like to be able to configure the width/number of characters per column (independently?) in browse view etc. so that I can adjust space according to preference and avoid content outside viewable area.

**Describe alternatives you've considered**
Perhaps also adjusted settings for automatic column condensation depending on content and window size. Certain columns sometimes have unused space while other columns are too narrow or outside window area.

aome510:
The clipped `duration` column seems to be because not all columns in the track table are constructed using percentage constraint:

https://github.com/aome510/spotify-player/blob/3b2a36625ac5978e8e68cb662bad340e15cc29a4/spotify_player/src/ui/page.rs#L743-L750

Regarding the adjustable column size, do you mean able to configure the percentage of each column? For example,
```toml
title = 25
artists = 30
```

Autobeing:
Precisely, something like that is what I had in mind. Percentages would make it easy to understand, with a clear default value as well.

cyberfeign:
Yep, that looks good. Odd that it's still clipping, maybe due to how the width of the window is set?

apprehensions:
Please! this damn padding at the bottom of my spotify_player has been bothering me for too long

PR #482, opened 2024-06-30
Add ability to adjust window ratios 

### LINKED ISSUE #480: Implementation of adjustable widget sizes
While working on implementing different sized widget in the library, I have come across two different way to implement the setting of different widget sizes in the config. 

### Option 1

Have two config values `album width` & `playlist width` and have them be the percentage values used in the layout. 

#### Pros 
- Super simple to understand

#### Cons 
- Having two values to configure the size of the widgets seems very unclean 
- It is easily possible to write values which are too large and would then cause errors, which need to be checked for 

### Option 2

Add a ration to the config such as `3:2:4`, which will define the size of the different widgets. See: [CSS-Flex ](https://developer.mozilla.org/en-US/docs/Web/CSS/flex)

#### Pros 
- One value to define the size of all relevant widgets
- Errors can be entirely avoided 
- Easier to scale to different places in the app 

#### Cons 
- A tiny bit harder to understand 
- Will require adding a new "type" to the config

DOD-101:
resolve #480

This PR adds the following: 

1. A new `layout` section to the config which contains options for adjusting the ratios of windows found within the app
2. Move `playback_window_width` (now `playback_window_height`) & `playback_window_position` to new `layout` section
3. Update docs accordingly

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -7,6 +7,7 @@
   - [Media control](#media-control)
   - [Player event hook command](#player-event-hook-command)
   - [Device configurations](#device-configurations)
+  - [Layout configurations](#layout-configurations)
 - [Themes](#themes)
   - [Use script to add theme](#use-script-to-add-theme)
   - [Palette](#palette)
@@ -14,7 +15,7 @@
 - [Keymaps](#keymaps)
 
 All configuration files should be placed inside the application's configuration folder (default to be `$HOME/.config/spotify-player`).
-
+    
 ## General
 
 **The default `app.toml` can be found in the example [`app.toml`](../examples/app.toml) file.**
@@ -139,6 +140,23 @@ The configuration options for the [Librespot](https://github.com/librespot-org/l
 
 More details on the above configuration options can be found under the [Librespot wiki page](https://github.com/librespot-org/librespot/wiki/Options).
 
+### Layout configurations
+
+The layout of the application can be adjusted via these options. 
+
+| Option                     | Description                                              | Default |
+| -------------------------- | -------------------------------------------------------- | ------- |
+| `library.album_percent`    | Adjust the width of the album window in the library      | `40`    |
+| `library.playlist_percent` | Adjust the width of the playlist window in the library   | `40`    |
+
+Example: 
+
+``` toml
+
+[layout]
+library = { album_percent = 40, playlist_percent = 40 }
+
+```
 ## Themes
 
 `spotify_player` uses the `theme.toml` config file to look for user-defined themes.
--- a/docs/config.md
+++ b/docs/config.md
@@ -27,7 +27,7 @@ All configuration files should be placed inside the application's configuration 
 | `client_id`                       | the Spotify client's ID                                                                  | `65b708073fc0480ea92a077233ca87bd`                      |
 | `client_port`                     | the port that the application's client is running on to handle CLI commands              | `8080`                                                  |
 | `tracks_playback_limit`           | the limit for the number of tracks played in a **tracks** playback                       | `50`                                                    |
-| `playback_format`                 | the format of the text in the playback's window                                          | `{status} {track} • {artists}\n{album}\n{metadata}`              |
+| `playback_format`                 | the format of the text in the playback's window                                          | `{status} {track} • {artists}\n{album}\n{metadata}`     | 
 | `notify_format`                   | the format of a notification (`notify` feature only)                                     | `{ summary = "{track} • {artists}", body = "{album}" }` |
 | `notify_timeout_in_secs`          | the timeout (in seconds) of a notification (`notify` feature only)                       | `0` (no timeout)                                        |
 | `player_event_hook_command`       | the hook command executed when there is a new player event                               | `None`                                                  |
@@ -43,9 +43,9 @@ All configuration files should be placed inside the application's configuration 
 | `enable_cover_image_cache`        | store album's cover images in the cache folder                                           | `true`                                                  |
 | `notify_streaming_only`           | only send notification when streaming is enabled (`streaming` and `notify` feature only) | `false`                                                 |
 | `default_device`                  | the default device to connect to on startup if no playing device found                   | `spotify-player`                                        |
-| `play_icon`                       | the icon to indicate playing state of a Spotify item                                     | `▶`                                                    |
+| `play_icon`                       | the icon to indicate playing state of a Spotify item                                     | `▶`                                                     |
 | `pause_icon`                      | the icon to indicate pause state of a Spotify item                                       | `▌▌`                                                    |
-| `liked_icon`                      | the icon to indicate the liked state of a song                                           | `♥`                                                    |
+| `liked_icon`                      | the icon to indicate the liked state of a song                                           | `♥`                                                     |
 | `border_type`                     | the type of the application's borders                                                    | `Plain`                                                 |
 | `progress_bar_type`               | the type of the playback progress bar                                                    | `Rectangle`                                             |
 | `playback_window_position`        | the position of the playback window                                                      | `Top`                                                   |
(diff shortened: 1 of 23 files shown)


spotify_player/src/ui/page.rs
```diff
@@ -326,12 +328,25 @@ pub fn render_library_page(
     // - a playlists window
     // - a saved albums window
     // - a followed artists window
+
+    let playlist_window_width: u16;
+    let album_window_width: u16;
+
+    if configs.app_config.playlist_window_width + configs.app_config.album_window_width > 99 {
```
aome510, line 335: I don't think this is needed. We should return an error if user specifies incorrect parameters. Otherwise, they won't know if they did something wrong

docs/config.md
```diff
@@ -53,6 +53,8 @@ All configuration files should be placed inside the application's configuration
 | `cover_img_length`                | the length of the cover image (`image` feature only)                                     | `9`                                                     |
 | `cover_img_scale`                 | the scale of the cover image (`image` feature only)                                      | `1.0`                                                   |
 | `seek_duration_secs`              | the duration (in seconds) to seek when using `SeekForward` and `SeekBackward` commands   | `5`                                                     |
+| `playlist_window_width`           | the width of the playlist window in the library                                          | `40`                                                    |
+| `album_window_width`              | the width of the album window in the library                                             | `40`                                                    |
```
aome510, line 57: can we make it clear that the value represents percentage w.r.t the whole application, e.g both in naming and documentation?

aome510:
I don't think this PR is related to https://github.com/aome510/spotify-player/issues/278. 278 asks for adjustable columns in a track table specifically not windows in the library page.

docs/config.md
aome510: for example, `library_playlist_width_percent` and `library_album_width_percent`

aome510: A much better way will be to add separate config called `layout`. `layout` contains configs for components like `library_page`, `track_table`, etc. Then `library_page` can consists of `album_percent`, `playlist_percent`, etc

DOD-101: When you say add a "separate config" do you mean add it's own `layout.toml`?

aome510: No it can still be inside `AppConfig` https://github.com/aome510/spotify-player/blob/ed9c4d476909b0755c5b1902b504f5ec6dd125ef/spotify_player/src/config/mod.rs#L48

the inside `app.toml` you can define `layout` configs under `[layout]` section. For example,

```toml
[layout]
library_page = { album_percent = 50, playlist_percent = 30 }
track_table = { album_percent = 30, track_percent = 20 }
```


One benefit of separate config is that you can also add config option to change horizontal/vertical layout for each component.

spotify_player/src/ui/page.rs
```diff
@@ -666,9 +679,9 @@ pub fn render_queue_page(
             .collect::<Vec<_>>(),
         [
             Constraint::Percentage(5),
-            Constraint::Percentage(40),
-            Constraint::Percentage(35),
-            Constraint::Percentage(20),
+            Constraint::Percentage(5),
+            Constraint::Percentage(5),
+            Constraint::Percentage(85),
```
aome510, line 684: this should be reverted

docs/config.md
```diff
@@ -137,6 +143,30 @@ The configuration options for the [Librespot](https://github.com/librespot-org/l
...
+The layout of the application can be adjusted via these options. 
+
+| Option                     | Description                                                   | Default |
+| -------------------------- | ------------------------------------------------------------- | ------- |
+| `library.album_percent`    | The width of the album window in the library                  | `40`    |
+| `library.playlist_percent` | The width of the playlist window in the library               | `40`    |
+| `search.top_percent`       | The height of the first row of search results                 | `50`    |
+| `search.left_percent`      | The width of the left column of search results                | `50`    |
+| `artist.album_percent`     | The width of the album window on the artist page              | `50`    |
+| `artist.top_song_percent`  | The height of the top songs window on the artist page         | `60`    |
```
aome510, line 157: let's try to make the scope of this PR small to **only include layout for library**. I have some problems with how these values are defined but we should leave that to a follow-up PR.

spotify_player/src/config/mod.rs
```diff
@@ -303,14 +331,98 @@ impl Default for DeviceConfig {
...
+}
+
+impl CheckValues for ArtistLayoutConfig {
+    fn check_values(&self) -> Result<(), String> {
+        if self.album_percent > 99 || self.top_songs_percent > 99 {
+            return Err("Artist-Layout options are invalid!".to_string());
+        }
+        Ok(())
+    }
+}
```
aome510, line 410: this introduces a bit of code boilerplates. You can implement all the checks inside a single `Layout::check_values` function, e.g

```
if self.album_percent + self.playlist_percent > 99 {
     anyhow::bail!("invalid library layout: summation of album_percent and playlist_percent cannot be greater than 99!");
}
```

aome510: the `check_values` function can return [`anyhow::Error`](https://docs.rs/anyhow/latest/anyhow/struct.Error.html) instead

spotify_player/src/ui/page.rs
```diff
@@ -47,15 +48,21 @@ pub fn render_search_page(
...
+    .iter()
+    .flat_map(|rect| {
+        Layout::horizontal([
+            Constraint::Percentage(configs.app_config.layout.search.left_percent),
+            Constraint::Percentage(100 - configs.app_config.layout.search.left_percent),
+        ])
+        .split(*rect)
+        .to_vec()
+    })
+    .collect::<Vec<_>>();
```
aome510, line 65: let's not make this change in this PR

spotify_player/src/ui/page.rs
```diff
@@ -713,13 +726,22 @@ fn render_artist_context_page_windows(
...
+    .split(rect);
     let top_tracks_rect = chunks[0];
 
     // albums and related artitsts windows
-    let chunks = Layout::horizontal([Constraint::Ratio(1, 2); 2]).split(chunks[1]);
+    let chunks = Layout::horizontal([
+        Constraint::Percentage(configs.app_config.layout.artist.album_percent),
+        Constraint::Percentage(100 - configs.app_config.layout.artist.album_percent),
+    ])
+    .split(chunks[1]);
```
aome510, line 744: let's not make this change in this PR

aome510 (changes requested):

DOD-101:
I have made all the changes you have requested. 

I apologize for the amount of problems with the code. I am new to rust and contributing to large public projects.

docs/config.md
```diff
@@ -115,7 +114,6 @@ set -euo pipefail
 
 case "$1" in
     "Changed") echo "command: $1, old_track_id: $2, new_track_id: $3" >> /tmp/log.txt ;;
-    "Playing") echo "command: $1, track_id: $2, position_ms: $3, duration_ms: $4" >> /tmp/log.txt ;;
```
aome510, line 118: did you delete this accidentally?

spotify_player/src/config/mod.rs
```diff
@@ -303,14 +313,43 @@ impl Default for DeviceConfig {
...
+                album_percent: 40,
+            },
+            playback_window_position: Position::Top,
+            playback_window_height: 6,
+        }
+    }
+}
+
+impl LayoutConfig {
+    fn check_values(&self) -> Result<(), anyhow::Error> {
```
aome510, line 330:
```suggestion
    fn check_values(&self) -> anyhow::Result<()> {
```

spotify_player/src/config/mod.rs
```diff
@@ -303,14 +313,43 @@ impl Default for DeviceConfig {
...
         }
 
-        Ok(config)
+        match config.layout.check_values() {
+            Err(e) => {
+                println!("{}", e);
+                std::process::exit(1)
+            }
+            Ok(_) => Ok(config),
+        }
```
aome510, line 352:
```suggestion
        config.layout.check_values()?;
        Ok(config)
```

aome510 (commented):
Thanks for updating. Only a few comments left. Can you also update the default values in https://github.com/aome510/spotify-player/blob/master/examples/app.toml as well?

aome510:
> I apologize for the amount of problems with the code. I am new to rust and contributing to large public projects.

No need to apologize lol. Instead, I appreciate that you spend your own time learning a new (famously hard) language and contributing to this project 🙇

docs/config.md
DOD-101: Yup. Fixed in d5d18c1.

aome510 (approved):

PR #501, opened 2024-07-10
add jump to track command

### LINKED ISSUE #484: Jump to currently playing track in the list
For those who have more than 1000 tracks in the playlist and use shuffle, it would be nice to have an action to navigate to the currently playing track in the current context. This action can be bound to the key sequence `g c` or `g 0`. Actually in Vim there's `z z` to center the current line under the cursor, but I see `z` already bound to queue.

 - If nothing is currently playing -> do nothing
 - If the current context doesn't contain a track -> either do nothing, or call `CurrentlyPlayingContextPage` action before jumping.

As an alternative, I suggest adding a way to jump to the track by it's number. Possibly by pressing the `#` key and typing in a track number. This also requires adding `{trackno}` to the playback format configuration, so the user can look at the track number, press `#` (or `g space` to switch to a context and then `#`) and jump to that track.

aNNiMON:
Resolves #484

I decided to refresh my ancient Rust knowledge and resolve my own feature request by myself 😄

It turns out that the simple jump to track action is better suited for this feature than the alternative of pressing `#` and entering a track number. The latter requires a track position to be persistent in the model, while currently track numbers are calculated during rendering.

### DIFF
--- a/spotify_player/src/event/mod.rs
+++ b/spotify_player/src/event/mod.rs
@@ -593,6 +593,25 @@ fn handle_global_command(
                 current_field: PlaylistCreateCurrentField::Name,
             });
         }
+        Command::JumpToCurrentTrackInContext => {
+            state.player.read()
+                .current_playing_track()
+                .and_then(|track| track.id.clone())
+                .and_then(|track_id| match ui.current_page() {
+                    PageState::Context { id: Some(context_id), .. } => {
+                        state.data.read().caches.context.get(&context_id.uri())
+                            .and_then(|context| match context {
+                                Context::Tracks { tracks, .. } => {
+                                    tracks.iter().position(|t| t.id == track_id)
+                                }
+                                _ => None,
+                            })
+                    }
+                    _ => None,
+                })
+                .map(|i| ui.current_page_mut().select(i));
+            ui.popup = None;
+        }
         Command::ClosePopup => {
             ui.popup = None;
         }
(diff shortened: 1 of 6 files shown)


spotify_player/src/event/mod.rs
```diff
@@ -220,6 +220,10 @@ pub fn handle_action_in_context(
                 }
                 ui.popup = None;
             }
+            Action::JumpToTrack => {
+                client_pub.send(ClientRequest::JumpToTrack(track.id))?;
+                ui.popup = None;
```
aome510, line 225: no need to define additional `ClientRequest`, you can handle the action directly here. The `jump_to_track` implementation doesn't really need any API calls provided by the client.

aNNiMON: Done, thanks

aome510:
Thanks for updating @aNNiMON. As I think about this more, I think it's better to define the feature as a `Command` instead of a `Action`. The reason is that this feature is specific to the currently playing track while a track action should be generic. For example, it has no use when you call that action in an **already selected track** in the context.

Then the command can be named `JumpToCurrentTrackInContext` or something with a detailed description so that user can know how to use it.

aNNiMON:
@aome510 done. I'm using `g c` key map here (c — **c**urrent track). Let me know if this key map should be changed to something else (e.g. `g 0`).

PR #508, opened 2024-07-17
show album types in the artist context

aNNiMON:
If an artist has many singles it's hard to find an album in the list:
image

I added an album type information and converted the list widget to a table:
image

### DIFF
--- a/spotify_player/src/state/model.rs
+++ b/spotify_player/src/state/model.rs
@@ -1,6 +1,6 @@
 pub use rspotify::model as rspotify_model;
 use rspotify::model::CurrentPlaybackContext;
-pub use rspotify::model::{AlbumId, ArtistId, Id, PlaylistId, TrackId, UserId};
+pub use rspotify::model::{AlbumId, AlbumType, ArtistId, Id, PlaylistId, TrackId, UserId};
 
 use crate::utils::map_join;
 use html_escape::decode_html_entities;
@@ -137,6 +137,7 @@ pub struct Album {
     pub release_date: String,
     pub name: String,
     pub artists: Vec<Artist>,
+    pub album_type: String,
 }
 
 #[derive(Deserialize, Serialize, Debug, Clone)]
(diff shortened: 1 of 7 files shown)


spotify_player/src/state/ui/popup.rs
```diff
@@ -18,7 +18,7 @@ pub enum PopupState {
     DeviceList(ListState),
     ArtistList(ArtistPopupAction, Vec<Artist>, ListState),
     ThemeList(Vec<crate::config::Theme>, ListState),
-    ActionList(ActionListItem, ListState),
+    ActionList(Box<ActionListItem>, ListState),
```
aome510, line 21: Why does this need to be put into `Box`?

spotify_player/src/state/model.rs
```diff
@@ -334,6 +336,12 @@ impl From<rspotify_model::FullAlbum> for Album {
             id: album.id,
             release_date: album.release_date,
             artists: from_simplified_artists_to_artists(album.artists),
+            album_type: match album.album_type {
+                AlbumType::Album => "album".to_string(),
+                AlbumType::Single => "single".to_string(),
+                AlbumType::Compilation => "compilation".to_string(),
+                AlbumType::AppearsOn => "appears on".to_string(),
+            },
```
aome510, line 344: let's keep the original representation `Option<AlbumType>` in case we want to process the type in the future. You can implement `fn type(&self) -> String` method for `Album` that returns the type as string

spotify_player/src/state/ui/popup.rs
aome510: Oh I see. It was because of the clippy warning

spotify_player/src/state/model.rs
aNNiMON: Done

aome510:
`main` failed to build because of conflicts between two commits. @aNNiMON is this something you can take a look and fix?

aNNiMON:
> is this something you can take a look and fix? 

On it

PR #514, opened 2024-07-23
Use album id as hash to store album covers

### LINKED ISSUE #506: OS error: file name too long
**Describe the bug**
When playing a song from a playlist that contains very long authors or song name, spotify_player runs into problems attempting to save files that have names which exceed the (mac) OS length limit for writing.

**To Reproduce**
On MacOS sonoma, play a song from an album such as [this one](https://open.spotify.com/album/16zfntcR7VAJWn9Ovooo1w?si=D1L7rw4lTE2fbOsFq-fjvg). I did this on a separate chrome spotify client, before trying to launch the terminal. Now that the song is active, the terminal will produce this output:
```
$ spotify_player
Error: Failed to initialize the Spotify data

Caused by:
    File name too long (os error 63)
```

**Expected behaviour**
The CLI doesn't attempt to write files which have names that exceed the mac limit. Perhaps they should be clipped, or encoded, etc.

**Log and backtrace**
No backtrace necessary, (my backtrace is entirely empty, this is my first launch of spotify_player). Here is the relevant log context:
```
2024-07-14T17:17:50.192136Z  INFO spotify_player::streaming: Initializing a new integrated player with device_id=c563f912-f32a-42fe-b29e-b6c585ea8eda
2024-07-14T17:17:50.192260Z  INFO spotify_player::streaming: Starting an integrated Spotify player using librespot's spirc protocol
2024-07-14T17:17:50.542336Z  INFO spotify_player::client: Retrieving image from url: https://i.scdn.co/image/ab67616d0000b273bfd2db0a9adc1f642860d277
2024-07-14T17:17:50.714032Z  INFO spotify_player::client: Saving the retrieved image into /Users/user/.cache/spotify-player/image/Love Songs - Piano Transcriptions Without Words-Angela Hewitt, Franz Liszt, Robert Schumann, Franz Schubert, Richard Strauss, Christoph Willibald Gluck, Gottfried Heinrich Stölzel, Gustav Mahler, Edvard Grieg, Gabriel Fauré, Manuel de Falla, George Gershwin, Percy Grainger-cover.jpg
```

It's clear from above what is causing the long file error.

**Environment**
- OS: Macos Sonoma 14.5
- Application version: 0.18.2

leon332157:
Fixes #506 

My attempt to address #506, this has been tested working on my windows machine, I do recommend more testing though, because with a small digest size, hash collision could occur. In that case, we could use a multi-folder caching technique.
 
I'm also new to rust, so hopefully I didn't break anything else :)


leon332157:
There is another approach which is by using the album id as the hash input to avoid collision where the same songs with different album covers write to the same hash. I will attempt to implement that.

spotify_player/Cargo.toml
```diff
@@ -11,6 +11,7 @@ readme = "../README.md"
 
 [dependencies]
 anyhow = "1.0.86"
+blake3 = "1.5.3"
```
aome510, line 14: no need to use a 3rd party library, rust has https://doc.rust-lang.org/std/hash/index.html

spotify_player/src/client/mod.rs
```diff
@@ -1547,9 +1546,14 @@ impl Client {
     async fn retrieve_image(
         &self,
         url: &str,
-        path: &std::path::Path,
+        filename: &str,
         saved: bool,
     ) -> Result<Vec<u8>> {
+        let configs = config::get_config();
+        let hashedname = crate::utils::hash_filename(filename);
```
aome510, line 1553: you can append the hash at the end of the name instead. Also make it shorter, e.g use `6` characters instead of `16`? The main idea to keep the album's name and artist's name is to make it easy for users to find

aome510 (changes requested):

spotify_player/src/client/mod.rs
leon332157: I used 6 digits of the album id as the hash now

aome510 (approved):
Can you also update the PR's title?

PR #524, opened 2024-07-29
Does not return removed commands

### LINKED ISSUE #523: Remap / rebind keymap to action
**Describe the bug**
It does not seem possible to remap a kemap (like C-p which is bind to "SelectPreviousOrScrollUp") to an action (like "AddToPlaylist").

**To Reproduce**
I have this **keymap.toml**

```
[[keymaps]]
command = "SelectPreviousOrScrollUp"
key_sequence = "k"

[[actions]]
action = "AddToPlaylist"
key_sequence = "C-p"
```

**Expected behaviour**
SelectPreviousOrScrollUp is k
AddToPlaylist is C-p

**Environment**

- OS: Linux Fedora 40
- Application version: 0.19.1

arouene:
Commands set to None to remove the keybinding, where immediately returned by **find_command_or_action_from_key_sequence** on l.418.
Instead of looking for action keybinding.

Fix #523

### DIFF
--- a/spotify_player/src/config/keymap.rs
+++ b/spotify_player/src/config/keymap.rs
@@ -397,6 +397,7 @@ impl KeymapConfig {
         self.keymaps
             .iter()
             .find(|&keymap| keymap.key_sequence == *key_sequence)
+            .filter(|keymap| keymap.command != Command::None)
             .map(|keymap| keymap.command)
     }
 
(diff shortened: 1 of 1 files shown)


spotify_player/src/config/keymap.rs
```diff
@@ -397,6 +397,7 @@ impl KeymapConfig {
         self.keymaps
             .iter()
             .find(|&keymap| keymap.key_sequence == *key_sequence)
+            .filter(|keymap| keymap.command != Command::None)
```
aome510, line 400: you can include this condition in `find`, e.g `keymap.key_sequence == *key_sequence && keymap.command != Command::None`

arouene: It was a stylistic preference :D, but if you prefer the other way, it's also fine.

aome510 (approved):

PR #525, opened 2024-07-30
feat: Allow themable "like" component

arouene:
Allow a themable "like" component

image


aome510 (approved):
Can you update https://github.com/aome510/spotify-player/blob/master/docs/config.md#component-styles as well?

PR #529, opened 2024-07-31
fix #454 overflowing of song title

### LINKED ISSUE #454: Title overflows when too long
**Describe the bug**
When listening to a song with a long name, the name overflows to the the album cover & metadata

**To Reproduce**
Listen to a song with a long title, such as 't​á​lcunnana d​ë​hajma tun dejl b​ë​natsë abcul​’​han dlhenic ëlh inagat, jahadlhë adrhasha indauzgl​ë​n nu dlhevusao ibajngra nava l​í​eshtamhan ëf novejhan conetc dan​ë​ctc qin, ëf tu dlhicad​ë​trh​ë​nna bë ablhundrhaba judjenan alh​ë​tangrasë shidandlham​ë​së inqom' by Trhä

**Expected behaviour**
The song title doesn't overflow everywhere.

**Screenshots**
image

If listened to another song,
The previous album cover appears (no song played previously):

Default app.toml:
image

DOD-101:
~This PR needs to be merged after #482~ resolved

Fix #454 by splitting the metadata text and the "title" text into two
different rects. This means that the title text being cut off to long
won't cause the metadata text to overflow outside the viewport.

This does however mean that the `playback_format` config option should
no longer contain the `{metadata}` placeholder.

### DIFF
--- a/spotify_player/src/ui/playback.rs
+++ b/spotify_player/src/ui/playback.rs
@@ -102,7 +102,7 @@ pub fn render_playback_window(
 
             if let Some(ref playback) = player.buffered_playback {
                 let playback_text = construct_playback_text(ui, track, playback);
-                let playback_desc = Paragraph::new(playback_text).wrap(Wrap { trim: false });
+                let playback_desc = Paragraph::new(playback_text);
                 frame.render_widget(playback_desc, metadata_rect);
             }
 


aome510:
> This PR needs to be merged after https://github.com/aome510/spotify-player/pull/482

Why is this the case? I thought you can create an upstream PR from a branch in your fork.

spotify_player/src/config/mod.rs
```diff
@@ -348,6 +349,7 @@ impl Default for LayoutConfig {
             },
             playback_window_position: Position::Top,
             playback_window_height: 6,
+            title_height: 1,
```
DOD-101, line 352: This PR requires #482 because the config option is implemented in the new Layout section of the config, which is added in that PR.

But considering the extensive changes made/requested to t#482, it might be better to close this PR  until that one is merged.

DOD-101: I hope this answers your question.

aome510:
I thought the overflow can be fixed by increasing `playback_window_width`. I don't see why we need a separate `title_height`?

apprehensions:
The title should simply be truncated if it can't fit fully. The official spotify client 'wraps' the title by moving it from left to right, but i don't believe that will be possible here.

apprehensions:
Could you squash your commits?

aome510:
> The title should simply be truncated if it can't fit fully. The official spotify client 'wraps' the title by moving it from left to right, but i don't believe that will be possible here.

truncating can be done by disabling wrapping when rendering the playback info. 

https://github.com/aome510/spotify-player/blob/2030b4ee94d1f7a4b6cb64d7fa56aa5336f3414e/spotify_player/src/ui/playback.rs#L105

DOD-101:
> truncating can be done by disabling wrapping when rendering the playback info.

Yes, however this would mean also cutting off the metadata info and what's if some people would rather have the text wrap so they can actually read it. 

> I thought the overflow can be fixed by increasing playback_window_width. I don't see why we need a separate title_height?

The problem with this is that some people have their metadata at the bottom of the playback window. When the text now wraps the metadata disappears. 

image

vs

image

Adding a separate rect for the metadata would fix this. It would mean that the title / album / artist wrapping wouldn't affect the position of the metadata. 

> The title should simply be truncated if it can't fit fully.

This can be done by simply setting `title_height` to 1 (which is the default).

DOD-101:
> Could you squash your commits?

Yes, I can, but why? 

Since the commits get squashed when merged into main it makes seems to me there is no difference once the PR is merged. 

So why would you like me to squash them?

aome510:
> The problem with this is that some people have their metadata at the bottom of the playback window. When the text now wraps the metadata disappears.

I don't see how this can happen. The whole playback info is constructed as a single paragraph widget. Wrapping in one line shouldn't affect others. See the below picture, title line is wrapped but it doesn't affect other lines.

image

@apprehensions' problem in the original issue happened because of the wrapping as he used spaces to represent paddings between playback section and cover image section, which doesn't work with text wrap. My suggestion is just simply disabling wrapping, then no overflowing issue will happen.

However, I can see that text wrapping can be convenient feature for some people.

DOD-101:
> Wrapping in one line shouldn't affect others. 

I'm not entirely sure what you mean by "affect", but what happens to me is that when the title line wraps it now takes up two lines, which means the line for the metadata is now no longer visible. 

Here is my config for reference: 

``` toml
playback_format = """
{track} • {album} • {artists}



{metadata}"""
playback_window_height = 6
```

aome510:
> > Wrapping in one line shouldn't affect others.
> 
> I'm not entirely sure what you mean by "affect", but what happens to me is that when the title line wraps it now takes up two lines, which means the line for the metadata is now no longer visible, because it is pushed down a line.
> 
> Here is my config for reference:
> 
> ```toml
> playback_format = """
> {track} • {album} • {artists}
> 
> 
> 
> {metadata}"""
> playback_window_height = 6
> ```

Can you try replacing whitespaced with newline (`\n`)? Maybe whitespace doesn’t work well with text wrapping

DOD-101:
> Can you try replacing whitespace with newline (\n)? Maybe whitespace doesn’t work well with text wrapping

Same thing happens.

aome510:
> > Can you try replacing whitespace with newline (\n)? Maybe whitespace doesn’t work well with text wrapping
> 
> Same thing happens.
> 
> This is the new config.
> 
> ```toml
> playback_format = """{track} • {album} • {artists}\n\n\n\n{metadata}"""
> ```

Ah I see, it's because you add extra lines to push the metadata to the end. Therefore, it's expected for the metadata part to be moved outside the view upon overflowing.

aome510:
Let's disable wrapping then. It's much simpler and will also resolve your problem.

DOD-101:
> Ah I see, it's because you add extra lines to push the metadata to the end. 

Yup. 

> Let's disable wrapping then. It's much simpler and will also resolve your problem.

I can implement this if you like, but since we can achieve the same with setting the `title_height` to 1 as a default and give people the option to adjust it using the option. (Effectively toggling wrapping) It is the option I prefer / suggest.

aome510:
I don't like separating `metadata` and other sections because it adds extra complexity while being less flexible, e.g. what if user wants to render metadata before album, artist?

DOD-101:
Fair point. I'll go ahead and change the code to turn off wrapping.

aome510 (approved):

apprehensions:
While the title and album look like theyre simply truncated, there is no visual indicator such as `...`, additionally, the album cover is missing.

aome510:
>  While the title and album look like theyre simply truncated, there is no visual indicator such as `...`, additionally, the album cover is missing.

probably because it failed to save the image file when album's name is too long

PR #539, opened 2024-08-08
feat: target configuration for action keymaps

### LINKED ISSUE #428: Remap key to key combination
I'm thinking how to bind y to the equivalent of 'a 5'. So, current song, copy link.
I think either having it so you can bind to a sequence of commands, or bind a key to a sequence of keys would be nice.
Something like
[[keymaps]]
command = "ShowActionsOnCurrentTrack; CopyTrackLink"
key_sequence = "y"
or
[[keymaps]]
command = "a 5" // maybe a different name instead of command? idk
key_sequence = "y"

juliamertz:
Actually fixes #428 this time 🙃

### DIFF
--- a/spotify_player/src/config/keymap.rs
+++ b/spotify_player/src/config/keymap.rs
@@ -1,5 +1,5 @@
 use crate::{
-    command::{Action, Command, CommandOrAction},
+    command::{Action, ActionTarget, Command, CommandOrAction},
     key::{Key, KeySequence},
 };
 use anyhow::Result;
@@ -25,6 +25,7 @@ pub struct Keymap {
 /// A keymap that triggers an `Action` when a key sequence is pressed
 pub struct ActionMap {
     pub key_sequence: KeySequence,
+    pub target: Option<ActionTarget>,
     pub action: Action,
 }
 
@@ -401,11 +402,14 @@ impl KeymapConfig {
     }
 
     /// finds an action from a mapped key sequence
-    pub fn find_action_from_key_sequence(&self, key_sequence: &KeySequence) -> Option<Action> {
+    pub fn find_action_from_key_sequence(
+        &self,
+        key_sequence: &KeySequence,
+    ) -> Option<(Action, ActionTarget)> {
         self.actions
             .iter()
             .find(|&action| action.key_sequence == *key_sequence)
-            .map(|action| action.action)
+            .map(|action| (action.action, action.target.unwrap_or_default()))
     }
 
     /// finds a command or action from a mapped key sequence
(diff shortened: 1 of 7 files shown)


spotify_player/src/config/keymap.rs
```diff
@@ -25,6 +25,7 @@ pub struct Keymap {
 /// A keymap that triggers an `Action` when a key sequence is pressed
 pub struct ActionMap {
     pub key_sequence: KeySequence,
+    pub target: Option<ActionTarget>,
```
aome510, line 28:
```suggestion
     #[serde(default)]
    pub target: ActionTarget,
```

spotify_player/src/event/page.rs
```diff
@@ -194,7 +194,7 @@ fn handle_key_sequence_for_search_page(
                 CommandOrAction::Command(command) => {
                     window::handle_command_for_artist_list_window(command, artists, &data, ui)
                 }
-                CommandOrAction::Action(action) => {
+                CommandOrAction::Action(action, _) => {
```
aome510, line 197: should this be

```suggestion
                CommandOrAction::Action(action, ActionTarget::SelectedItem) => {
```
?

Same with other matches in this file

spotify_player/src/event/mod.rs
```diff
@@ -105,8 +108,13 @@ fn handle_key_event(
 
     // if the key sequence is not handled, let the global command handler handle it
```
aome510, line 109:
```suggestion
    // if the key sequence is not handled, let the global handler handle it
```

spotify_player/src/event/mod.rs
```diff
@@ -346,6 +383,40 @@ fn handle_show_actions_on_artist(
...
+        let data = state.data.read();
+
+        let currently_playing = match player.current_playing_track() {
+            Some(track) => track,
+            None => anyhow::bail!("No currently playing track"),
+        };
+        let track = match Track::try_from_full_track(currently_playing.clone()) {
+            Some(track) => track,
+            None => anyhow::bail!("No currently playing track"),
+        };
```
aome510, line 406: normally, we don't return error when there is nothing to act on when handling a key event, e.g. no playing track. See how `Command::ShowActionsOnCurrentTrack` is handled in `event/mod.rs`:

```rust
  if let Some(track) = state.player.read().current_playing_track() {
      if let Some(track) = Track::try_from_full_track(track.clone()) {
          let data = state.data.read();
          ...
```

aome510 (commented):
Thanks for the PR. I left a few comments other looks good.

spotify_player/src/event/page.rs
juliamertz: You're right, this was an oversight on my part

juliamertz:
Should all be fixed now, only formatting caused a conflict

aome510 (approved):

Issue #535, opened 2024-08-07
[Feature]: Ability to further customize layout


NickTheWilder:
**Describe the solution you'd like**
I personally never use Albums & Artists in Spotify, only playlist to sort songs. It would be nice to be able to 
1. Remove columns
2. Add different columns to replace them (seems more complicated but a nice to have)

**Describe alternatives you've considered**
Adjusting the layout of playlists/albums in the config to potentially hide the columns

ℹ️ FWIW the layout settings don't work for me, Win11 - PowerShell v7.4.4 - v0.19.1
Current settings: 
```toml
# omitted other settings above
[layout]
library = { playlist_percent = 75, album_percent = 12 }
playback_window_position = "Top"
```

**Additional context**
I'd start working on this if I had experience with Rust and open a PR but no experience here. Really enjoy what's going on so far.

Example to further describe what I'm talking about

Current view:
![image](https://github.com/user-attachments/assets/00cc9e86-5cce-4152-a525-d2815cc2b4a9)

Proposed view: (or smth similar)
![image](https://github.com/user-attachments/assets/7c2e5107-27dc-4ded-8d59-9c06e8dfc669)

apprehensions:
Would love to see this, it bridges the gap between the official spotify client and the better TUI counterpart.

DOD-101:
I'm looking into implementing this. 

> FWIW the layout settings don't work for me, Win11 - PowerShell v7.4.4 - scoop v0.19.1

That's strange. Are you on the latest git?

arvvnd:
The playback don't seem to work for me either, but in my case I tried to change the `playback_window_position = Bottom`.
I use kitty on arch and I used the AUR package `spotify-player-full-pipe`.

inside my app.toml :
```[layout]
library = { playlist_percent = 40, album_percent = 40 }
playback_window_position = "Bottom"
#playback_position = "Bottom"
playback_window_height = 6
#playback_window_height = 12
```

I launch spotify_player in kitty using the command `spotify_player` and connect to my own client ID. I am able to edit other settings on the app.toml file like theme, icons, etc., and observe these changes. However, when I edit the [layout] settings, nothing changes. I even changed `playback_window_height` from its default value of 6 to 12 to see if the [layout] settings would take effect, but the playback window size remained the same. After making changes to app.toml, I save the file, restart kitty, and open the player with the spotify_player command to check for changes.

[...]

PR #545, opened 2024-08-14
feat(layout): minor adjustment to layout config

DOD-101:
This addresses the first request in #535.  

> 1. Remove columns

The change is mainly in adjusting how the borders are drawn.

It doesn't address the second request:

> Add different columns to replace them 

It might also be a good idea to implement some guard clauses to prevent the unseen windows from being rendered etc.

### DIFF
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -328,8 +328,8 @@ impl Default for LayoutConfig {
 
 impl LayoutConfig {
     fn check_values(&self) -> anyhow::Result<()> {
-        if self.library.album_percent + self.library.playlist_percent > 99 {
-            anyhow::bail!("Invalid library layout: summation of album_percent and playlist_percent cannot be greater than 99!");
+        if self.library.album_percent + self.library.playlist_percent > 100 {
+            anyhow::bail!("Invalid library layout: summation of album_percent and playlist_percent cannot be greater than 100!");
         } else {
             Ok(())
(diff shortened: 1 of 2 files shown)


NickTheWilder:
(New to rust so bear with me)

I'd like to test this since it's a feature I requested to at least contribute what I can. How could I do that? 
Is there already a contributing/how-to-build-locally doc that I'm missing?

I run `cargo b` to build the application, and then `cargo r` to run and I get the errors:

image

Cargo version: `cargo 1.80.1 (376290515 2024-07-16)`
Git: `2.46.0 (latest version)`

DOD-101:
That's strange. There shouldn't be any errors. (Especially since all the checks are passing)

Try running `git diff` / `git status` to see if something may have accidental changed. 

> Is there already a contributing/how-to-build-locally doc that I'm missing?

Not that I know of.

Tip: You can just run cargo r(un). It will build and then run.

aome510 (commented):
I don't think increasing other windows' size to "remove" a column is a good idea. The column is not really removed, it's just hidden from the UI. `FocusNextWindow` still points to the "hidden" window, which is confusing.

DOD-101:
Do you think we should have a separate config option for this, or do we have an internal check if any column has a width of 0 and then skip any code relating to?
 
I think the option of an internal check would be better, since it's less config for the user to write.

DOD-101:
Closing in favor of #567

PR #548, opened 2024-08-16
feat: client_id_cmd config option

### LINKED ISSUE #537: FR: Store `client_id` outside dotfiles
I do version control and publish my `.config` directory in a GitHub repo, as quite a lot of people do. Thus, Due to the inclusion of the `client_id`, this results in having to gitignore the `~/.config/spotify_player/app.toml`, as otherwise the id would be leaked.

Save `client_id` outside the `app.toml`, so the remainder of the spotify-player-config can be stored in a public git repo without any trouble.

The client_id could then be read from an environment variable, which one could save in `.zshenv`.

Alternatively, support something like the `password_cmd` from `spotifyd`: https://docs.spotifyd.rs/config/File.html
```toml
# This is an alternative to the `password` field. Both
# can't be used simultaneously.
password_cmd = "command_that_writes_password_to_stdout"
```

- gitignoring the `app.toml` and manually updating it all the time.
- symlinking the `app.toml` into a private cloud storage.

juliamertz:
This PR fixes #537 by adding a new config option `client_id_cmd`. When set, it will run this command and use its stdout as the client_id. I haven't been able to verify it's working on Windows, but that part has been copied from the rust docs so it should be fine.

### DIFF
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -48,6 +48,7 @@ impl Configs {
 pub struct AppConfig {
     pub theme: String,
     pub client_id: String,
+    pub client_id_cmd: Option<String>,
 
     pub client_port: u16,
 
@@ -229,6 +230,7 @@ impl Default for AppConfig {
             theme: "dracula".to_owned(),
             // official Spotify web app's client id
             client_id: "65b708073fc0480ea92a077233ca87bd".to_string(),
+            client_id_cmd: None,
 
             client_port: 8080,
 
@@ -385,6 +387,34 @@ impl AppConfig {
             ap_port: self.ap_port,
             ..Default::default()
         }
+    }
+
+    /// Returns stdout of `client_id_cmd` if set, otherwise it returns the the value of `client_id`
+    pub fn get_client_id(&self) -> Result<String> {
+        if let Some(cmd) = self.client_id_cmd.clone() {
+            let output = if cfg!(target_os = "windows") {
+                std::process::Command::new("cmd")
+                    .args(["/C", &cmd])
+                    .output()
+            } else {
+                std::process::Command::new("sh")
+                    .arg("-c")
+                    .arg(&cmd)
+                    .output()
+            }?;
+
+            if !output.status.success() {
+                let stderr = String::from_utf8(output.stderr)?;
+                tracing::error!("Running command: {}, returned error: {}", cmd, stderr);
+                anyhow::bail!(stderr);
+            }
+
+            let stdout = String::from_utf8(output.stdout)?;
+            tracing::info!("Running command: {}, returned client id: {}", cmd, stdout);
+            return Ok(stdout);
+        }
+
+        Ok(self.client_id.clone())
     }
 }
 
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -143,15 +143,15 @@ pub struct Command {
 }
 
 impl Command {
+     /// Execute a command, returning stdout if succeeded or stderr if failed
     pub fn execute(&self, extra_args: Option<Vec<String>>) -> anyhow::Result<String> {
         let mut args = self.args.clone();
-        extra_args.map_or_else(|| {}, |extra| args.extend(extra));
+        args.extend(extra.unwrap_or_default());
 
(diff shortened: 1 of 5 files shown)


spotify_player/src/config/mod.rs
```diff
@@ -386,6 +388,34 @@ impl AppConfig {
...
+            let output = if cfg!(target_os = "windows") {
+                std::process::Command::new("cmd")
+                    .args(["/C", &cmd])
+                    .output()
+            } else {
+                std::process::Command::new("sh")
+                    .arg("-c")
+                    .arg(&cmd)
+                    .output()
+            }?;
```
aome510, line 404: Should we use `config::Command` struct to represent this for consistency (e.g. with previous `player_event_hook_command`)? See https://github.com/aome510/spotify-player/blob/c3e9a4e8b30e00240fb262c89c25de43eb195bdc/spotify_player/src/config/mod.rs#L138-L142

I guess specifying a command string is more convenient, but the above code looks kinda "hacky" to me.

juliamertz: Ah I didn't know this existed in the codebase, this does seem like a better idea. 
I could pull out the execution logic for `player_event_hook_command` and implement that on the Command struct

spotify_player/src/streaming.rs
```diff
@@ -143,18 +143,8 @@ fn execute_player_event_hook_command(
...
-        .args(&args)
-        .output()?;
-
-    // running the player hook command failed, report the command's stderr as an error
-    if !output.status.success() {
-        let stderr = std::str::from_utf8(&output.stderr)?.to_string();
-        anyhow::bail!(stderr);
-    }
+    let args = Some(event.args());
+    cmd.execute(args)?;
```
aome510, line 147:
```suggestion
    cmd.execute(Some(event.args()))?;
```

spotify_player/src/config/mod.rs
```diff
@@ -386,6 +408,14 @@ impl AppConfig {
             ..Default::default()
         }
     }
+
+    /// Returns stdout of `client_id_command` if set, otherwise it returns the the value of `client_id`
+    pub fn get_client_id(&self) -> Result<String> {
+        match self.client_id_command.clone() {
+            Some(cmd) => cmd.execute(None),
+            None => Ok(self.client_id.clone()),
+        }
```
aome510, line 417:
```suggestion
        match self.client_id_command {
            Some(ref cmd) => cmd.execute(None),
            None => Ok(self.client_id.clone()),
        }
```

spotify_player/src/config/mod.rs
```diff
@@ -141,6 +142,26 @@ pub struct Command {
...
+impl Command {
+    pub fn execute(&self, extra_args: Option<Vec<String>>) -> anyhow::Result<String> {
+        let mut args = self.args.clone();
+        extra_args.map_or_else(|| {}, |extra| args.extend(extra));
+
+        let output = std::process::Command::new(&self.command)
+            .args(&args)
+            .output()?;
+
+        // running the command failed, report the command's stderr as an error
```
aome510, line 154:
```suggestion
```

spotify_player/src/config/mod.rs
```diff
@@ -141,6 +142,26 @@ pub struct Command {
     pub args: Vec<String>,
 }
 
+impl Command {
+    pub fn execute(&self, extra_args: Option<Vec<String>>) -> anyhow::Result<String> {
```
aome510, line 146:
```suggestion
     /// Execute a command, returning stdout if succeeded or stderr if failed
    pub fn execute(&self, extra_args: Option<Vec<String>>) -> anyhow::Result<String> {
```

spotify_player/src/config/mod.rs
```diff
@@ -141,6 +142,26 @@ pub struct Command {
...
+        let mut args = self.args.clone();
+        extra_args.map_or_else(|| {}, |extra| args.extend(extra));
```
aome510, line 148:
```suggestion
        args.extend(extra.unwrap_or_default());
```

aome510 (approved):
few comments left. Otherwise, looks good

juliamertz:
I learn new things about this language every day, thanks!

PR #549, opened 2024-08-18
Update README.md

ManiArasteh:
I updated README.md with important notice (Problem when authenticating). If it was good, merge it. otherwise close it without merging


aome510 (approved):
Thanks for adding the notice

PR #559, opened 2024-09-04
Update layout of library and search pages for vertical aspect ratios

juliamertz:
I've chosen a columns to rows ratio that works well for me, but I'm not sure if this is the best approach since terminal rendering varies.


aome510 (approved):
Sorry for the late review and thank you for the PR. I tested it. Really cool feature!

juliamertz:
No worries, thank you for taking the time!

PR #562, opened 2024-09-07
Fix proposal for #561

### LINKED ISSUE #561: The 'notify' feature requires the 'streaming' feature, otherwise it won't compile
**Describe the bug**
See title: Building the app with the 'notify' feature set also requires the 'streaming' feature to be set, otherwise it won't compile.

**To Reproduce**
Build the app with: `cargo build --release --bin spotify_player --no-default-features --features notify`

**Expected behaviour**
It should be possible to compile 'notify' on its own (or make it dependent on 'streaming').

The problem lies in the if statement in spotify_player/src/client/mod.rs:1464 : It gets enabled ONLY by the 'notify' feature flag but in ln1465 two fields are accessed which are only present when BOTH 'notify' and 'streaming' flags are set (see spotify_player/src/client/mod.rs:36 and spotify_player/src/config/mod.rs:109). Thus it won't compile.

As I'm not familiar with the project I'm not sure if it is fixed by simply changing client/mod.rs:ln1463 from  #[cfg(feature = "notify")] to #[cfg(all(feature = "streaming", feature = "notify"))] so that  these fields are only accessed when we are sure that they are present. But I suspect this would somewhat break the notify feature (if compiled on its own). 

**Log and backtrace**
Compiler error:
```
error[E0609]: no field `notify_streaming_only` on type `AppConfig`
    --> spotify_player/src/client/mod.rs:1465:37
     |
1465 |             && (!configs.app_config.notify_streaming_only || self.stream_conn.lock().is_some())
     |                                     ^^^^^^^^^^^^^^^^^^^^^ unknown field
     |
     = note: available fields are: `theme`, `client_id`, `client_id_command`, `client_port`, `player_event_hook_command` ... and 20 others

error[E0609]: no field `stream_conn` on type `&client::Client`
    --> spotify_player/src/client/mod.rs:1465:67
     |
1465 |             && (!configs.app_config.notify_streaming_only || self.stream_conn.lock().is_some())
     |                                                                   ^^^^^^^^^^^ unknown field
     |
     = note: available fields are: `http`, `spotify`, `auth_config`
     = note: available field is: `session`

For more information about this error, try `rustc --explain E0609`.
error: could not compile `spotify_player` (bin "spotify_player") due to 2 previous errors
```

**Screenshots**

**Environment**

- OS: Opensuse Tumbleweed
- Application version: cargo 1.81.0, spotify-player (latest master)
- Application features: any feature combination with 'notify' and without 'streaming'

**Additional context**
Maybe a good first issue? I'd like to fix it / open a PR by myself but I'm quite unfamiliar with this project (and with rust).

DJ-pH3luy:
Fix conditional compilation that it works with notify enabled and steaming disabled. See #561 for details.

Fixes #561

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1468,8 +1468,7 @@ impl Client {
         }
 
         #[cfg(all(feature = "notify", not(feature = "streaming")))]
-        if configs.app_config.enable_notify
-        {
+        if configs.app_config.enable_notify {
             Self::notify_new_track(track, &path)?;
         }
 
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1460,9 +1460,15 @@ impl Client {
         }
 
         // notify user about the playback's change if any
-        #[cfg(feature = "notify")]
+        #[cfg(all(feature = "notify", feature = "streaming"))]
         if configs.app_config.enable_notify
             && (!configs.app_config.notify_streaming_only || self.stream_conn.lock().is_some())
+        {
+            Self::notify_new_track(track, &path)?;
+        }
+
+        #[cfg(all(feature = "notify", not(feature = "streaming")))]
+        if configs.app_config.enable_notify
         {
             Self::notify_new_track(track, &path)?;
         }


aome510 (approved):
thanks for the fix

PR #572, opened 2024-10-09
Update README.md to include a hint for WSL

ShuraShulgin:
added a little bit of text which hints WSL users to run `cargo install spotify_player` with `--locked` so it'll work.
This should avoid further issues being created to find this solution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,11 +91,7 @@ Run `scoop install spotify-player` to install the application.
 
 ### Cargo
 
-Run `cargo install spotify_player` to install the application from [crates.io](https://crates.io/crates/spotify_player).
-
-#### Notes for WSL
-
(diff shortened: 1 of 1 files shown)


aome510 (changes requested):
This error is not specific to `wsl`. It's better to update https://github.com/aome510/spotify-player?tab=readme-ov-file#cargo to include `--locked` instead and add a note saying that `--locked` is important for cargo to use the packaged `Cargo. lock` file

ShuraShulgin:
(I rarely work with github, I hope I did not press any annoying buttons, If I did, I'm very sorry)

aome510 (approved):

aome510:
> (I rarely work with github, I hope I did not press any annoying buttons, If I did, I'm very sorry)

all good, thanks for the PR!

PR #576, opened 2024-10-14
feat(layout): extend layout functionality

NickTheWilder:
Starts the first part of #535 to hide specific rectangles based on the user's configuration file. 
Inspired from the review comments in #545 

Opened as a draft bc I've limited out on knowledge of the application (and rust).

TODO: 
- [ ] Set the focused state to the first 'found' rectangle.
- [x] Update default configuration file.
- [x] Extend README.

Feel free to takeover and create a new branch or directly add commits here.

### DIFF
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -191,8 +191,9 @@ pub struct LayoutConfig {
 
 #[derive(Debug, Deserialize, Serialize, ConfigParse, Clone)]
 pub struct LibraryLayoutConfig {
-    pub playlist_percent: u16,
-    pub album_percent: u16,
+    pub playlist_percent: Option<u16>,
+    pub album_percent: Option<u16>,
+    pub artist_percent: Option<u16>,
 }
 
 #[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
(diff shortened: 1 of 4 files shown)


NickTheWilder:
I see that the blob below sets the default focus state to the playlists, how can I import the config to this and set the initial focused state to the first 'found' percentage in the config?

https://github.com/aome510/spotify-player/blob/8dcbc225cafd179b84b261b56ab1620dfcac1fc9/spotify_player/src/state/ui/page.rs#L231-L241

--

My thought was smth like

```rust
if let Some(value) = configs.app_config.layout.library.playlist_percent {
    return LibraryFocusState::Playlists;
} else if let Some(value) = configs.app_config.layout.library.album_percent {
    return LibraryFocusState::SavedAlbums;
} else if let Some(value) = configs.app_config.layout.library.artist_percent {
    return LibraryFocusState::FollowedArtists;
}
```

spotify_player/src/config/mod.rs
```diff
@@ -361,8 +383,17 @@ impl LayoutConfig {
...
+            // If a config file was found and parsed, reset the library percentages
+            // to ensure only the user-specified values are set
+            config.layout.library.playlist_percent = None;
+            config.layout.library.album_percent = None;
+            config.layout.library.artist_percent = None;
+
+            // Re-parse the config file to set only the user-specified values
+            config.parse_config_file(path)?;
+        } else {
+            config.write_config_file(path)?;
```
NickTheWilder, line 396: this could be moved out of if statement

PR #598, opened 2024-11-02
Use 'ring' backend as default for rustls.

### LINKED ISSUE #596: Use `ring` as default crypto provider with rustls
librespot has already switched to not using rustls' new default crypto provider aws-lc-sys as of https://github.com/kingosticks/librespot/commit/24bbc6314c991c64d1846162f739f223c44e5779 .  

spotify-player should follow.  aws-lc does not even work on a lot of platforms.

With this diff, I still get the aws-lc-sys dependency pulled in on OpenBSD, but at least the application works now.

```
diff --git a/spotify_player/Cargo.toml b/spotify_player/Cargo.toml
index 75678d2..cd4f29c 100644
--- a/spotify_player/Cargo.toml
+++ b/spotify_player/Cargo.toml
@@ -57,7 +57,7 @@ clap_complete = "4.5.35"
 which = "7.0.0"
 fuzzy-matcher = { version = "0.3.7", optional = true }
 html-escape = "0.2.13"
-rustls = "0.23.16"
+rustls = { version = "0.23.16", default-features = false, features = ["ring"] }
 
 [target.'cfg(any(target_os = "windows", target_os = "macos"))'.dependencies.winit]
 version = "0.30.5"
diff --git a/spotify_player/src/main.rs b/spotify_player/src/main.rs
index 4d00e99..2342900 100644
--- a/spotify_player/src/main.rs
+++ b/spotify_player/src/main.rs
@@ -221,7 +221,7 @@ async fn start_app(state: &state::SharedState) -> Result<()> {
 fn main() -> Result<()> {
     // librespot depends on hyper-rustls which requires a crypto provider to be set up.
     // TODO: see if this can be fixed upstream
-    rustls::crypto::aws_lc_rs::default_provider()
+    rustls::crypto::ring::default_provider()
         .install_default()
         .unwrap();
 

```

pstumpf:
Resolves #596

### DIFF
--- a/spotify_player/Cargo.toml
+++ b/spotify_player/Cargo.toml
@@ -57,7 +57,7 @@ clap_complete = "4.5.35"
 which = "7.0.0"
 fuzzy-matcher = { version = "0.3.7", optional = true }
 html-escape = "0.2.13"
-rustls = "0.23.16"
+rustls = { version = "0.23.16", default-features = false, features = ["ring"] }
 
 [target.'cfg(any(target_os = "windows", target_os = "macos"))'.dependencies.winit]
 version = "0.30.5"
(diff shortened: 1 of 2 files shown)


aome510 (approved):
Thank you!

PR #607, opened 2024-11-21
Changed binary (executable) file name in Cargo.toml

ManiProjs:
Hi! I changed Cargo.toml to save the output of binary with file name “sptplay”.  You can recommend another name, too. 
It makes running application easier.


aome510:
Hi, thanks for the PR but I don't have any plans of changing the binary name anytime soon

PR #34, opened 2022-05-08
Add `lyric_finder` lib crate

aome510:
Implemented the `lyric_finder` crate that retrieves a song's lyric using https://genius.com and its APIs

PR #622, opened 2024-12-01
Add native-tls-alpn feature to reqwest

ikopytko:
I encountered a problem similar to one described in #34.
For unknown reason Genius was blocking my requests for some time, response for the search request was:

> Sorry, you have been blocked
> You are unable to access genius.com
 
Both browser and curl sucessfully returned the expected json.
I came across [this discusion](https://github.com/seanmonstar/reqwest/discussions/2227#discussioncomment-9403885) in reqwest library and after trying the solution it lyric finder finally start working again.

### DIFF
--- a/lyric_finder/Cargo.toml
+++ b/lyric_finder/Cargo.toml
@@ -14,7 +14,7 @@ rustls-tls = ["reqwest/rustls-tls"]
 native-tls = ["reqwest/native-tls"]
 
 [dependencies]
-reqwest = { version = "0.12.4", features = ["json"], default-features = false }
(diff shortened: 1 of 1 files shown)


aome510 (approved):

cycle-five:
Could this have some relevance? https://stackoverflow.com/questions/74652786/how-to-set-apln-protocols-before-tls-handshake-with-openssl-in-rust 
I really would prefer to use both your library, and to not pull openssl into my project, so I'm willing to take on this issue myself but it's a blocker for me.

PR #632, opened 2024-12-08
Don't pull in openssl

cycle-five:
The feature added to reqwest I don't believe is needed and it unnecessarily pulls in openssl into the project.


aome510:
See https://github.com/aome510/spotify-player/pull/622

cycle-five:
Could this have some relevance? https://stackoverflow.com/questions/74652786/how-to-set-apln-protocols-before-tls-handshake-with-openssl-in-rust
I really would prefer to use both your library, and to not pull openssl into my project, so I'm willing to take on this issue myself but any help on where the issue can be fixed within the rustls feature branch would be appreciated.

aome510:
@cycle-five with #635 landed, the use `lyric-finder` in the main project is no longer needed. Feel free to fork the implementation yourself. It's pretty short. In the meantime, I'll close this PR for now. Lmk if you have any other questions.

PR #636, opened 2024-12-15
Cleanup `config::theme` codes

aome510:
- avoid panic on lyrics page upon no playback
- clean up codes in `config/theme.rs`


manh9203 (approved):
LGTM!

PR #658, opened 2025-01-06
fix: building with notify only

dibbel26:
While (#562) implemented conditional compilation for notify without streaming the used function seems to be deprecated with &track out of scope. Replacing it with Self::notify_new_track(&track, &path)?; fixes the issue.


aome510 (approved):
Thanks for fixing this!

PR #664, opened 2025-01-16
fix: Some shows not loading

### LINKED ISSUE #663: Crosslinked issue on Podcasts not loading
See https://github.com/ramsayleung/rspotify/issues/515

I think this can also be fixed without waiting for the fix to land in rspotify which moves pretty slowly judging by their release history?

m4r1vs:
fixes #663
Can be reverted once it is fixed upstream in rspotify

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1409,14 +1409,31 @@ impl Client {
         let show = self.get_a_show(show_id, None).await?;
         let first_page = show.episodes.clone();
 
+        // Copy first_page but use Page<Option<SimplifiedEpisode>> instead of Page<SimplifiedEpisode>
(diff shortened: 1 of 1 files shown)


m4r1vs:
Not much experience with rust so I'm not sure if there is a cleaner way to copy all the properties of first_page without doing it explicitly like this..

aome510:
> Not much experience with rust so I'm not sure if there is a cleaner way to copy all the properties of first_page without doing it explicitly like this..

Normally, you can do

```rust
       let first_page = rspotify::model::Page {
             items: first_page.items.into_iter().map(Some).collect(),
             ..first_page
         };
```

However, in this case, it's not possible because the type of `items` is changed from `Vec<T>` to `Vec<Option<T>>`

PR #678, opened 2025-02-07
Fixing bug when performing an action on an album

JohnOberhauser:
Fixing a bug where performing an action on an album can apply that action to the wrong album.

Introduced in my last PR https://github.com/aome510/spotify-player/pull/654

My original PR sorted the UI, but I didn't realize actions on albums happen based on the selected _index_ (not the selected album).  So I'm adding album sorting to the album event window as well.

### DIFF
--- a/spotify_player/src/event/window.rs
+++ b/spotify_player/src/event/window.rs
@@ -1,5 +1,6 @@
 use super::page::handle_navigation_command;
 use super::*;
+use crate::utils::sort_albums_by_type;
 use crate::{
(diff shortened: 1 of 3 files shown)


aome510:
I think it's better to sort the albums upon getting data from the API. This avoids having to sort the data each time we process it. Put up [a PR](https://github.com/aome510/spotify-player/pull/697)

JohnOberhauser:
Awesome, thanks!

Issue #684, opened 2025-03-01
Search returns json parse error


EliRibble:
**Describe the bug**
Any time I search I either get an error (direct search as a command argument) or nothing happens (type in search from search page)

**To Reproduce**
Start spotify-player. Hit "g s" to go to the search page. Type in any search query. Hit "enter". Nothing happens in the UI, all results stay empty.

Alternatively, run spotify_player with:

```
spotify_player search anything
```

Output is:

> Bad request: json parse error: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 4573: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 4573

**Expected behaviour**
To see search results, or output indicating no matches

**Log and backtrace**
```
2025-03-01T00:08:14.752929Z ERROR client_request{request=Search("anything")}: spotify_player::client::handlers: Failed to handle client request: json parse error: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 2140: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 2140
2025-03-01T00:08:14.862643Z  INFO client_request{request=GetCurrentUserQueue}: spotify_player::client: Successfully handled the client request, took: 334ms
2025-03-01T00:08:17.874728Z ERROR client_request{request=Search("anything")}: spotify_player::client::handlers: Failed to handle client request: json parse error: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 2140: invalid type: null, expected struct SimplifiedPlaylist at line 1 column 2140
```
**Screenshots**

**Environment**

- OS: NixOS 24.11.714614.c618e28f7025 
- Application version: spotify_player 0.20.1

EliRibble:
Looking around in the source code I found a reference to https://github.com/ramsayleung/rspotify/issues/459. Seems the data model for responses from Spotify is defined in a library rather than in spotify-player's code.

EliRibble:
Threw in the `shell.nix` I created while investigating this.

PR #685, opened 2025-03-01
Add working shell.nix

EliRibble:
This is useful for developers, like me, that use NixOS for reproducible developer environments. This will get everything you need to build the default feature set as of today.

Created this while investigating https://github.com/aome510/spotify-player/issues/684

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ cd /usr/pkgsrc/audio/spotify-player
 make install
 ```
 
+### NixOS
+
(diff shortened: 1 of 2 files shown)


quotequack (approved):
Really good for now but I will try to implement a flake or default.nix later but for now this so good work

juliamertz:
@quotequack I've already implemented a flake a while ago for the fork I keep. If you want you can refer to that [here](https://github.com/juliamertz/spotify-player/blob/dev/nix/flake.nix)

PR #687, opened 2025-03-04
Fixed an issue where the player would get stuck on one song after playing for about an hour or so

### LINKED ISSUE #603: Player got stuck in one song
**Describe the bug**
After using spotify-player for like 15 minutes listening a playlist the player gets stuck on a song, even if you select previous or next, it stays on the same song, if you try to place another song by selecting it directly, it does not play the one you have chosen.

**To Reproduce**
1. start listening
2. wait at least 15 minutes
3. notice that one song got stuck

**Expected behaviour**
continue playing without issues

**Log and backtrace**
[spotify-player-24-11-06-11-21.log](https://github.com/user-attachments/files/17651186/spotify-player-24-11-06-11-21.log)

**Screenshots**
[Video](https://files.catbox.moe/25zeyw.mp4)

**Environment**

- OS: Windows 11 (Pro) 23H2 22631.4391
- Application version: spotify_player 0.20.1

alexjyong:
Partially fixes https://github.com/aome510/spotify-player/issues/603 

Looks like we weren't loading the new auth token into the Session when it got refreshed. 

Unfortunately, this doesn't seem to be a complete fix, as I see the issue crop up _sometimes_, and the current page you are on the UI will have to be manually refreshed when the token expires (i.e. navigating to another page seems to help)

It _does_ however, reduce the frequency when it happens. It seems to be the most stable when you get a long playlist running and let it be for a bit.


aome510:
> Looks like we weren't loading the new auth token into the Session when it got refreshed.

I don't think this is the case. I haven't been following the original issue but I'm pretty sure that `librespot` session implements a token refetching logic itself. The auth token that `spotify_player` implements refetching is used by the Spotify client to make API requests not the session

alexjyong:
@aome510 yeah I'm not 100% sure. I noticed the frequency of the "stuck in one song" issues went down a bit, but not completely. I posted some logs in the original issue.

aome510:
> @aome510 yeah I'm not 100% sure. I noticed the frequency of the "stuck in one song" issues went down a bit, but not completely. I posted some logs in the original issue.

Thank you for investigating! I'll take a deeper look when I get more time

alexjyong:
Hi @aome510 , any updates on this from your end? Just curious :)

aome510:
> Hi @aome510 , any updates on this from your end? Just curious :)

Sorry, I don't have any update. Couldn't really reproduce the issue on my end.

landnthrn:
That must of been FUN doing an hour per test 😆

PR #715, opened 2025-05-09
Add # j and # k to jump # number up and down.

markgandolfo:
This pull request adds the ability to mimic vim's j & k commands. 

Now we you can do `[1-9] j` or `[1-9] k` to move up or down that many lines. 

i.e. you can `4 j` to go down 4 lines. or `3 k` to go up 3 lines.

### DIFF
--- a/spotify_player/src/config/keymap.rs
+++ b/spotify_player/src/config/keymap.rs
@@ -32,299 +32,322 @@ pub struct ActionMap {
 
 impl Default for KeymapConfig {
     fn default() -> Self {
+        // Generate keymaps for `j` and `k` keys with a count in front
+        // i.e. 4 j -> SelectNextOrScrollDown { count: 4 }
+        fn generate_keymap_for_up_and_down() -> Vec<Keymap> {
+            let mut keymaps = vec![];
+
+            for i in 1..=9 {
+                keymaps.push(Keymap {
+                    key_sequence: format!("{i} j").as_str().into(),
+                    command: Command::SelectNextOrScrollDown { count: i },
+                });
+                keymaps.push(Keymap {
+                    key_sequence: format!("{i} k").as_str().into(),
+                    command: Command::SelectPreviousOrScrollUp { count: i },
+                });
+            }
+
+            keymaps
+        }
+
+        let mut keymaps = vec![
+            Keymap {
+                key_sequence: "n".into(),
+                command: Command::NextTrack,
+            },
+            Keymap {
+                key_sequence: "p".into(),
+                command: Command::PreviousTrack,
+            },
+            Keymap {
+                key_sequence: ".".into(),
+                command: Command::PlayRandom,
+            },
+            Keymap {
+                key_sequence: "space".into(),
+                command: Command::ResumePause,
+            },
+            Keymap {
+                key_sequence: "C-r".into(),
+                command: Command::Repeat,
+            },
+            Keymap {
+                key_sequence: "M-r".into(),
+                command: Command::ToggleFakeTrackRepeatMode,
+            },
+            Keymap {
+                key_sequence: "C-s".into(),
+                command: Command::Shuffle,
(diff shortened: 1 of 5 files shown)


spotify_player/src/event/page.rs
```diff
@@ -501,15 +501,15 @@ pub fn handle_navigation_command(
 
     let configs = config::get_config();
     match command {
-        Command::SelectNextOrScrollDown => {
-            if id + 1 < len {
-                page.select(id + 1);
+        Command::SelectNextOrScrollDown { count } => {
+            if id + count < len {
+                page.select(id + count);
```
aome510, line 506: should this be `page.select(min(id + count, len - 1))`? I assume that will be more user friendly

spotify_player/src/event/page.rs
```diff
@@ -501,15 +501,15 @@ pub fn handle_navigation_command(
...
             }
             true
         }
-        Command::SelectPreviousOrScrollUp => {
-            if id > 0 {
-                page.select(id - 1);
+        Command::SelectPreviousOrScrollUp { count } => {
+            if id - count > 0 {
+                page.select(id - count);
```
aome510, line 512: same here

spotify_player/src/event/popup.rs
```diff
@@ -436,16 +436,20 @@ fn handle_command_for_list_popup(
...
-        Command::SelectNextOrScrollDown => {
-            if current_id + 1 < n_items {
-                popup.list_select(Some(current_id + 1));
-                on_select_func(ui, current_id + 1);
+        Command::SelectNextOrScrollDown { count } => {
+            let step = count.min(n_items.saturating_sub(current_id + 1));
+            if step > 0 {
+                let new_id = current_id + step;
+                popup.list_select(Some(new_id));
+                on_select_func(ui, new_id)
```
aome510, line 452: why does this have different handling logic than the one implemented in `spotify_player/src/event/page.rs`?

spotify_player/src/config/keymap.rs
```diff
@@ -32,299 +32,322 @@ pub struct ActionMap {
...
+                    command: Command::SelectNextOrScrollDown { count: i },
+                });
+                keymaps.push(Keymap {
+                    key_sequence: format!("{i} k").as_str().into(),
+                    command: Command::SelectPreviousOrScrollUp { count: i },
+                });
+            }
+
+            keymaps
+        }
```
aome510, line 52: this doesn't have to be a function, you can add the logic as-is after defining `keymaps` variable

aome510 (commented):
The approach looks good in general but it seems to add too much verbose to the Command Help page:

image

Can you fix that? Ideally, we want to see something similar to what is shown on README

aome510:
@markgandolfo I plan to release a new version this weekend. If you don't have time to address my comments by then, it's fine, we can include this feature in the next release

markgandolfo:
> The approach looks good in general but it seems to add too much verbose to the Command Help page:
> 
> image
> Can you fix that? Ideally, we want to see something similar to what is shown on README

Yeah, that's not ideal at all.  let me have a look at how I can do that..

markgandolfo:
> @markgandolfo I plan to release a new version this weekend. If you don't have time to address my comments by then, it's fine, we can include this feature in the next release

nah i won't make it this weekend, but let me implement your suggestions and resubmit. 
I'm fairly new to rust, so thanks for the feedback

aome510:
> nah i won't make it this weekend, but let me implement your suggestions and resubmit.

take ur time. No rush

markgandolfo:
Hey @aome510 I'm rethinking the implementation. 
I'm going to close this, let me reimplement. I'm thinking rather than map 1-9j/k I'll abstract it so we can use it for other things like volume control, etc. 

i.e. 5+ could turn the volume up 5 increments.

PR #716, opened 2025-05-10
feat: display duration on albums and playlists

DOD-101:
You can now see the duration of albums and playlists when viewing them. 

I didn't add this to the `tracks` context, since I'm assuming that is for the liked songs page and could pose a performance issue.
Additionally, this is what the official Spotify app does as well.

image


aome510 (approved):
nice

PR #717, opened 2025-05-10
ci: fix ci failure

DOD-101:
Minor changes to make CI pass.

This also removes the need for `once_cel`as a dependency. (At least a direct one)


aome510 (approved):
Thank you for fixing the CIs!

PR #722, opened 2025-05-14
recommend official Arch package instead of AUR

ThomasFrans:
An official Arch package has been available since 2025-02-05. Since this is the simplest way to install packages on Arch, it makes sense for most users to prefer this over the AUR packages. The AUR package that is currently recommended in the documentation is also no longer available because of the adoption in the official repositories.


ThomasFrans:
https://archlinux.org/packages/extra/x86_64/spotify-player/

aome510 (approved):

masroof-maindak:
@ThomasFrans, would you by any chance know what features this official package is built with?

PR #727, opened 2025-05-17
docs: add rhel/fedora systems dependencies documentation

gumaath:
This PR adds commands to the documentation for installing the necessary dependencies on RHEL/Fedora-based systems using the `yum` or `dnf` package managers.


gumaath:
I will open it again with the proper branch name.

PR #755, opened 2025-06-24
Show relative time that a track was added in playlists

Funtimes909:
Uses the chrono_humanize crate, there may be a way to do it without that dependency but I didn't look into it too much, if adding a new dependency is a problem I can properly look into this.

Simply adds a new column to show the relative time that a track was added to a playlist


aome510 (commented):
nice UI addition. There are 2 things that I noticed
* with the new `Added` column, the `Duration` column is not fully visible even when there is a big enough space for the app's window:
Screenshot 2025-07-26 at 5 08 15 PM

* this doesn't work for some playlists, e.g. ones that are generated by Spotify. `55 years ago` is shown
image

aome510:
Another thing I noticed from the official app is that it displays relative time if less than some threshold (a month?) and absolute time otherwise. It would be nicer if we implement the same thing

Screenshot 2025-07-27 at 6 04 38 PM

Funtimes909:
Currently a bit busy IRL at the moment but will get to this when I have the time, thanks for looking over this!

aome510:
> Currently a bit busy IRL at the moment but will get to this when I have the time, thanks for looking over this!

Take your time. Thank you for contributing on this project in your free time, appreciate it!

PR #769, opened 2025-07-08
Fixing an issue where if the user built the application from source, the image art wouldn't load.

alexjyong:
I noticed when I built `spotify_player` from source, that the album cover wouldn't load. 

Did some updates to ``spotify_player/Cargo.toml`` and `Cargo.lock` and now the image art loads as expected.

Not too sure how this is working with official releases, as I didn't see anything in the cd job that was enabling this.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,9 +1,7 @@
 name: CD
 
 on:
-  push:
-    tags:
-      - "v*.*.*"
+  workflow_dispatch:
 
 jobs:
   publish:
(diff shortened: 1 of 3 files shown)


aome510:
> I noticed when I built spotify_player from source, that the album cover wouldn't load.

This is intended because `image` doesn't work well for some terminals, so it's an optional configuration. See [image section of the README](https://github.com/aome510/spotify-player?tab=readme-ov-file#image)

alexjyong:
@aome510 I completely missed that. my bad.

PR #771, opened 2025-07-11
fix: SearchPage, Fuzzy search, and playlist creation will now allow users to type in numbers

alexjyong:
Fixes: https://github.com/aome510/spotify-player/issues/764


aome510:
Thanks for the PR and sorry for the late reply. The above issue should be fixed by https://github.com/aome510/spotify-player/pull/754

PR #783, opened 2025-07-28
Allow configuring the log directory

### LINKED ISSUE #774: Disable or change the location of log files
Please go through the [Features README page](https://github.com/aome510/spotify-player?tab=readme-ov-file#features) to make sure your request feature hasn't been implemented.

It isn't.

A clear and concise description of what the problem is.

`spotify_player` creates log and backtrace files in its cache directory each time it's started. These files pile up quickly and need to be manually removed.

Some kind of CLI flag/config option to move and/or disable log files. I know that `--cache-folder` exists, but that moves the entire cache, including the credentials.

It would also be nice if the logs were stored in a temporary directory by default or at least cleaned up every once in a while.

acuteenvy:
New config option in `app.toml`: `log_folder`
This allows the user to change the location of log and backtrace files. The default is still the same (cache directory).

When `RUST_LOG` is set to `off`, the log and backtrace files won't be created.

Closes #774.

### DIFF
--- a/spotify_player/src/config/mod.rs
+++ b/spotify_player/src/config/mod.rs
@@ -37,7 +37,7 @@ pub struct Configs {
 impl Configs {
     pub fn new(config_folder: &std::path::Path, cache_folder: &std::path::Path) -> Result<Self> {
         Ok(Self {
-            app_config: AppConfig::new(config_folder)?,
+            app_config: AppConfig::new(config_folder, cache_folder)?,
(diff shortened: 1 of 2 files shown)


acuteenvy:
Could the default log directory be changed to something like `$XDG_RUNTIME_DIR/spotify-player`? There's no use for old log files in the cache.

aome510 (approved):

PR #805, opened 2025-08-19
Feature: Add ability to change position of progress bar

rvnku:
image
The progress bar can now be placed to the right of the album cover instead of under it. To set the position in the app.toml file, specify the Bottom (default) or Right value for the progress_bar_position parameter.


aome510 (approved):

sasspaper:
image

Is there a way to shrink the progress bar or at least make it go a bit to the right? My artwork is covering a part of it and this is the best I could come out with after tweaking with playback_window_height

PR #836, opened 2025-09-21
changes for fast navigation

Jwsxs:


spotify_player/src/config/keymap.rs
```diff
@@ -236,6 +236,10 @@ impl Default for KeymapConfig {
                     key_sequence: "down".into(),
                     command: Command::SelectNextOrScrollDown,
                 },
+                Keymap {
+                    key_sequence: "C-down".into(),
+                    command: Command::SelectNextOrScrollDown10,
```
aome510, line 241: why don't you just bind this key sequence to `PageSelectNextOrScrollDown` will do scroll down of `page_size_in_rows` (default to be `20`) instead of adding a new command

Jwsxs: wasn't sure on how to check the Ctrl command if not that way... also it's 10 for better visibility on doing so, but yeah I'll check that out

PR #841, opened 2025-09-29
Read client_id from a file in the config folder

rashil2000:
Related: #537 

This PR introduces a default lookup for a file containing the client_id, eliminating the need to add an external command to read it from the filesystem. 

It is a plaintext file called `client_id` present alongside app.toml (in the ~/.config/spotify-player) directory, which contains the ID - if such a file exists we can read it first, and then fallback to other mechanisms. 

My main motivation for this change is that shelling out to an external command is an expensive operation, and not all systems have straightforward ways to do that (i.e. Windows does not have the `cat` command, although there are ways to achieve similar functionality via the `findstr` or `more` commands). Of course this is assuming the commands are being invoked without a shell (where stuff like `echo` is available). But external shell -> external command is even more expensive.

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,7 @@ docker run --rm \
 
 To enable a full [Spotify connect](https://www.spotify.com/us/connect/) support, users will need to enable a _"user-provided client integration"_.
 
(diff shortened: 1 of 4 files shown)


rashil2000:
@aome510 I have another idea in mind: instead of an environment variable, what if there's a plaintext file called `client_id` present alongside app.toml (in the ~/.config/spotify-player) directory, which contains the ID - if such a file exists we can read it first, and then fallback to other mechanisms. I can change the implementation in this PR to do that. 

My main motivation for this change is that shelling out to an external command is an expensive operation, and not all systems have straightforward ways to do that (i.e. Windows does not have the `cat` command, although there are ways to achieve similar functionality via the `findstr` or `more` commands). Of course this is assuming the commands are being invoked without a shell (where stuff like `echo` is available). But external shell -> external command is even more expensive.

PR #875, opened 2025-11-16
feat(nix): add nix flake

DOD-101:
This way the package can be build and run using nix commands and the flake can be used to get a more up-to-date version of spotify-player than nixpkgs has. This can be particularly useful to nix users if there is a fix on master which hasn't made its way downstream.

`default.nix` has been taken directly from nixpkgs with just the maintainers and the passthru update-script removed.

I have also removed `shell.nix` and made that part of the flake directly. 

The additions to `.gitignore` are so that users of (nix-)direnv can add a `.envrc` containing `use flake` to load the dev shell.

### DIFF
--- a/checklist.md
+++ b/checklist.md
@@ -2,7 +2,7 @@
 
 ## Pre-release checklist
 
-- [ ] change the `version` value defined in `main.rs`, `Cargo.toml` and `default.nix`.
+- [ ] change the `version` value defined in `main.rs` and `Cargo.toml`.
 - [ ] run `cargo clippy` to check the codes as well as to update `Cargo.lock`
 - [ ] run `cargo publish` in the `spotify_player` folder to publish the package
 - [ ] create a new release in [`github`](https://github.com/aome510/spotify-player/releases/new)
--- a/default.nix
+++ b/default.nix
@@ -45,7 +45,11 @@ assert lib.assertOneOf "withAudioBackend" withAudioBackend [
 
 rustPlatform.buildRustPackage rec {
   pname = "spotify-player";
-  version = "0.21.2";
+  version =
+    let
+      toml = builtins.fromTOML (builtins.readFile ./spotify_player/Cargo.toml);
+    in
+    toml.package.version;
 
   src = ./.;
 


ApprenticeofEnder:
Genuinely needed upgrade here.

DOD-101:
This wouldn't change anything about how spotify-player is packaged and release in [nixpkgs](https://github.com/NixOS/nixpkgs). That would remain the main way for Nix(Os) users to get the package. 

What adding this flake would do is:

1. Allow users to add an "input" to their system `flake.nix` to get spotify-player directly from this repo, rather than waiting until upstream updates. You can think of it kind of like using a dependency directly from git in rust, rather than getting it from the crates registry. 

For an example of adding an input, see [my config.](https://github.com/DOD-101/NixOS-dots/blob/master/flake.nix)

2. (more minor) Makes development for Nix(Os) users slightly more comfortable, since they can now add a `.envrc` file to the directory, which contains `use_flake`. This will then automatically load a dev-shell to work in and then once they are done making their changes they can directly run `nix build .` to check if everything still works for Nix(Os)


----

I do understand the worry of added maintenance burden on your part. 

There are 2 scenarios (that I can think of) where this would occur:

1. Most obviously, if the `default.nix`, used to build the package, breaks. In this case, it needs to be fixed or more realistically copied over from nixpkgs again and slightly adjusted. (like I have done)

2. You would always need ~to change the version number in `default.nix` when changing versions and~ take into account the implications for the flake more generally when making large changes to spotify-player. 

The first issue could be addressed at least in part by using [naersk](https://github.com/nix-community/naersk), which I could adjust the flake to use. 

More generally, I will say that I am more than happy to take over responsibility for all nix-related things within the repo. If that is something you would be ok with.

aome510:
> More generally, I will say that I am more than happy to take over responsibility for all nix-related things within the repo. If that is something you would be ok with.

Yep, I would prefer this if possible. Can you update this PR to remove nix-related things in the repo and update README accordingly to point to your stuff?

btw, thanks for the context and detailed info about nix

DOD-101:
> Can you update this PR to remove nix-related things in the repo and update README accordingly to point to your stuff?

I can do that, but it would significantly negate the benefits of the flake. Since the dev-shell would be very difficult to use and (more importantly) there is no way for my flake in my repo to automatically update when this repo changes, so I would need to set up some complicated CI to auto-check for updates. 

What I was trying to say is that if you have any Issue / PR to do with nix, you could label it as `nix` (or similar) and then ping me, and I'd take the time to review it.

aome510:
> > Can you update this PR to remove nix-related things in the repo and update README accordingly to point to your stuff?
> 
> I can do that, but it would significantly negate the benefits of the flake. Since the dev-shell would be very difficult to use and (more importantly) there is no way for my flake in my repo to automatically update when this repo changes, so I would need to set up some complicated CI to auto-check for updates.
> 
> What I was trying to say is that if you have any Issue / PR to do with nix, you could label it as `nix` (or similar) and then ping me, and I'd take the time to review it.

Oh I see, thanks for the clarification

adamzwakk:
Thanks for this, I was just writing my own but this is even better <3

Edit: I'm getting this compile error trying to use it

 > warning: sixel-sys@0.5.0: WITHOUT PNG
       > error: failed to run custom build command for `sixel-sys v0.5.0`
       >
       > Caused by:
       >   process didn't exit successfully: `/build/ldy4w3dbpn675a74pnxk6wiajjh0ad44-source/target/release/build/sixel-sys-b7f2d322f1b0f56d/build-script-build` (exit status: 101)
       >   --- stdout
       >   cargo::warning=WITHOUT PNG
       >   running: cd "/build/cargo-vendor-dir/sixel-sys-0.5.0/libsixel" && "sh" "-c" "exec \"$0\" \"$@\"" "autoreconf" "-ivf"
       >
       >   --- stderr
       >   autoreconf: line 1: exec: autoreconf: not found
       >
       >   thread 'main' panicked at /build/cargo-vendor-dir/autotools-0.2.7/src/lib.rs:790:5:
       >
       >   command did not execute successfully, got: exit status: 127
       >
       >   build script failed, must exit now
       >   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

PR #900, opened 2025-12-25
feature: toast popup when you queue

chris90483:
Description of the feature: When you add a track to the queue (with shift + z), the application adds it to your queue silently. This PR changes this behavior by making a "toast" popup appear when a track gets added to the queue. The toast popup dismisses itself automatically after some time (I implemented a default duration in the app config).

You can see a little demo here: https://youtu.be/W4txYkqaxIc


mgumz:
@chris90483 maybe you can provide a screenshot and a description of the feature you drafted?

chris90483:
Hi @mgumz, thanks for your reply. The feature now works as intended so this PR is no longer in draft.

Description of the feature: When you add a track to the queue (with shift + z), the application adds it to your queue silently. This PR changes this behavior by making a "toast" popup appear when a track gets added to the queue. The toast popup dismisses itself automatically after some time (I implemented a default duration in the app config).

You can see a little demo here: https://youtu.be/W4txYkqaxIc

chris90483:
I'll keep this updated with the upstream every now and then. 

Feel free to download my fork if you want this feature :p

PR #905, opened 2025-12-30
fix(nix): add missing nativeBuildInputs

DOD-101:
These dependencies were missing after #891

Fixes: https://github.com/aome510/spotify-player/pull/875#issuecomment-3699803008

My bad for not rebasing #875 after the pushes to master.


DOD-101:
The ci should prevent this from happening again, but if you'd rather not add it, I can remove it again.

BerriJ:
I can confirm that this PR works :)

Issue #924, opened 2026-02-10
Strange terminal behaviour after 0.22 update


felprangel:
**Describe the bug**
Just upgraded to 0.22 and opening the app makes some strange behaviour on the terminal

My terminal is alacritty, tested on kitty and im facing the same error

The inputs in the image are from my mouse movement, any key pressed on the keyboard is also logged

Cntrl + C stops working too

**To Reproduce**
Just open the app after the new update

**Expected behaviour**
Expected to not open the terminal after the app launches

**Log and backtrace**
Backtrace is empty
Log:
```
[...]

PR #925, opened 2026-02-10
fix: keep main thread alive on Linux and when media-control is disabled - Issue 924

Douglas019BR:
Fix [Issue 924](https://github.com/aome510/spotify-player/issues/924)

The main thread was exiting prematurely, killing all spawned threads (UI, event handler, player watcher, etc.).

- On Linux with media-control enabled, add a polling loop since there is no winit event loop to block the main thread (unlike macOS/Windows).
- Without the media-control feature, add a polling loop for non-daemon mode to prevent the process from terminating while the UI is running.

### DIFF
--- a/spotify_player/src/main.rs
+++ b/spotify_player/src/main.rs
@@ -207,31 +207,12 @@ async fn start_app(state: &state::SharedState) -> Result<()> {
             #[allow(deprecated)]
             event_loop.run(move |_, _| {})?;
         }
-
-        // On Linux, we need to keep the main thread alive while the UI is running
-        #[cfg(not(any(target_os = "macos", target_os = "windows")))]
-        {
-            loop {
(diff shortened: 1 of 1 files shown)


Douglas019BR:
I have tested the change only in my environment (Ubuntu). Please test it on Windows and macOS to ensure that it hasn't caused any issues.

aome510:
> I have tested the change only in my environment (Ubuntu). Please test it on Windows and macOS to ensure that it hasn't caused any issues.

Thanks for the PR but I think it's still better to go back to previous behaviour: joining all running tasks at the end. Lemme put up the change

aome510:
actually, some of the spawned threads have infinite loop, so we can just add an infinite loop at the end of `start_app` to simplify it

Douglas019BR:
Your solution is much simpler.

PR #931, opened 2026-02-13
feat: add mouse scroll to incease/decrease volume

### LINKED ISSUE #446: Be able to bind keymaps to mouse scroll

**No**

No, it is not.

I would like be able to bind the volume to scroll, hence allowing the keymaps to be able to listen / match mouse scroll inputs.

```
[[keymaps]]
command = "VolumeUp"
key_sequence = "ScrollDown"

[[keymaps]]
command = "VolumeDown"
key_sequence = "ScrollUp"
```

Using my mouses software I theoretically could add a unused keys to the scroll directions, then use keymaps with those keys.

**Additional context**

This can be done on the default Spotify client by hovering the volume bar and scrolling.

Alimedhat000:
Adds mouse scroll support to increase/decrease volume.

Resolves #446.

In `app.toml`:
```
enable_mouse_scroll_volume = true # enable mouse scrolling to change volume
volume_scroll_step = 5  # percent to change per scroll (default)
```

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -64,6 +64,8 @@ A sample `app.toml` is available at [examples/app.toml](../examples/app.toml).
 | `cover_img_pixels`                | Pixels per side for cover image (requires `pixelate` feature).                                 | `16`                                                                   |
 | `seek_duration_secs`              | Seek duration in seconds for seek commands.                                                    | `5`                                                                    |
 | `sort_artist_albums_by_type`      | Sort albums by type on artist pages.                                                           | `false`                                                                |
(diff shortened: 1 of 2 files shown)


Alimedhat000:
would changing it to use a bool like `enable_mouse_scroll_volume` instead of the keymap option be better?
cause it's a scrolling option why would some one remap that?

Alimedhat000:
one problem i faced is that i had to reuse some of the of `VolumeChange` logic because it's not accessible from the mouse handler.

A potential fix would be to add a small helper function that the mouse handler could use.
i'm happy to implement this but i'd wait for your approval.

PR #942, opened 2026-03-02
Fix Spotify API February 2026 migration issues (v0.23.0)

Knightsmarian:
Fixes critical UI and API breakages caused by the February 2026 Spotify Web API migration for Development Mode apps. Migrates to /items endpoint, adds model aliases/defaults for stripped fields, and improves search resilience.

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -665,6 +665,85 @@ impl AppClient {
         Ok(self.device().await?)
     }
 
+    /// Get a show data
+    pub async fn get_a_show(&self, show_id: ShowId<'_>, market: Option<rspotify::model::Market>) -> Result<serde_json::Value> {
+        let url = format!("{SPOTIFY_API_ENDPOINT}/shows/{}", show_id.id());
+        let mut query = Query::new();
+        if let Some(market) = market {
+            query.insert("market", <&str>::from(market));
+        }
+        self.http_get::<serde_json::Value>(&url, &query).await
+    }
+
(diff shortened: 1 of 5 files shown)


Knightsmarian:
IMPORTANT: THIS IS JUST AN EMERGENCY FIX TO GET THE APP WORKING. 

Big items: fixed new api endpoints and parsing, added rate limits to stop getting error 429

Pages still broken: 
Albums
Artist
Playlists
Radios
Image Previews

These likely all stem from the same parsing issue with the new api 

You can get around most of these with your liked library ( g y ) and search ( g s )

aome510:
Based on https://developer.spotify.com/documentation/web-api/tutorials/february-2026-migration-guide,

> Extended Quota Mode apps: No migration required. Apps in extended quota mode are not affected by any of the changes described in this guide — all existing endpoints, fields, and behaviors remain unchanged. You can optionally adopt the new generic library endpoints (PUT/DELETE /me/library) but your existing integrations will remain fully functional.
>
> Development Mode apps: This guide is for you. Read on for the full details of what's changing and how to update your app.

so it seems to affect dev-mode apps i.e when user specifies their own `client_id` instead of using the default `NCSPOT_CLIENT_ID` as in https://github.com/aome510/spotify-player/blob/6cdcb784dace48f4e6fc8834d1515e3401352239/spotify_player/src/config/mod.rs#L281-L288

FWIW, I don't think the API changes should be handled by `spotify_player` but `rspotify` (see https://github.com/ramsayleung/rspotify/issues/550). `spotify_player` will handle the API integration layer not  API request construction layer which should be handled by `rspotify`

PR #943, opened 2026-03-06
Feat: Adding real-time audio visualisation

### LINKED ISSUE #689: Visualization feature
Music Visualization would be cool.

Spotify-Tui has an analysis feature that shows the levels of different notes. A similar feature to leave on as a pleasant background would be cool

Narqulie:
Resolves #689

### DIFF
--- a/spotify_player/src/ui/playback.rs
+++ b/spotify_player/src/ui/playback.rs
@@ -478,6 +478,7 @@ fn render_playback_cover_image(state: &SharedState, ui: &mut UIStateGuard) -> Re
 
 /// Split the given area into two, the first one for the playback window
 /// and the second one for the main application's layout (popup, page, etc).
+#[allow(unused_variables)]
 fn split_rect_for_playback_window(state: &SharedState, rect: Rect) -> (Rect, Rect) {
     let configs = config::get_config();
     let playback_width = configs.app_config.layout.playback_window_height;


Narqulie:
Screen Recording 2026-03-06 at 17 56 33

  Implements the frequency visualization requested in [#689.](https://github.com/aome510/spotify-player/issues/689)

  What this does

  Adds an opt-in bar chart to the playback window showing real-time frequency content of the music being played — 64 log-scale bands from bass (left) to treble (right), with a smooth decay so bars fade out gradually rather
  than snapping to silence.

  How it works

  Rather than using the deprecated Spotify Audio Analysis API, this taps directly into the librespot audio stream. A VisualizationSink wraps the real audio backend, intercepts decoded PCM samples, and runs an FFT using rustfft
  (https://crates.io/crates/rustfft). Results are written to a shared buffer that the UI reads each render frame.

  Key details:

   - Sliding window FFT (FFT_SIZE=1024, HOP_SIZE=256) gives ~5.8ms update rate — well ahead of the ~32ms UI refresh
   - Hann windowing to reduce spectral leakage
   - Bars rise instantly on loud transients, decay slowly with a configurable factor
   - Only active with the streaming feature; no-op and zero overhead otherwise
   - Disabled by default — opt in via enable_audio_visualization = true in app.toml

  To try it

  Add to your ~/.config/spotify-player/app.toml:

   enable_audio_visualization = true

  Then build with the streaming feature enabled (the default).

PR #966, opened 2026-04-07
Add confirmation popups on destructive actions

### LINKED ISSUE #960: Playlist Edit / Delete confirmation step
No, now track deletes without confirmation

No, its not a kind of bug, but very needed feature

When you try to delete / edit track or playlist a popup shows, y / n to delete or cancel.

sakashimaa:
Closes #960

Now on every destructive action in app we show popup with y/n, following / unfollowing and like / unlike track is not considered as a destructive actions. Screenshot:

image

### DIFF
--- a/spotify_player/src/event/mod.rs
+++ b/spotify_player/src/event/mod.rs
@@ -290,7 +290,7 @@ pub fn handle_action_in_context(
                 } = ui.current_page()
                 {
                     ui.popup = Some(PopupState::ConfirmAction {
-                        message: "Are you sure?".to_string(),
+                        message: "Delete this track?".to_string(),
                         action: ConfirmableAction::DeleteTrackFromPlaylist {
                             playlist_id: playlist_id.clone_static(),
                             track_id: track.id,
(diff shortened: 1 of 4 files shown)


spotify_player/src/event/mod.rs
```diff
@@ -289,12 +289,14 @@ pub fn handle_action_in_context(
                     ..
                 } = ui.current_page()
                 {
-                    client_pub.send(ClientRequest::DeleteTrackFromPlaylist(
-                        playlist_id.clone_static(),
-                        track.id,
-                    ))?;
+                    ui.popup = Some(PopupState::ConfirmAction {
+                        message: "Are you sure?".to_string(),
```
aome510: message should be specific. For example, one may try to delete a track but they might accidentally trigger unfollowing artist

sakashimaa: check now, changed to more detailed like "Delete this playlist?"

aome510: is it possible to include the name to be even more specific? Like Remove "track" from this playlist, Unfollow "artist", Unlike "track", etc

sakashimaa: yep look now

aome510 (approved):

PR #967, opened 2026-04-07
feat: pause playback on app quit

### LINKED ISSUE #944: Pause playback on app quit
No, it is a minor feature request

When closing the TUI (q), playback should be paused. If a song is playing and i close the app, then reopen it, the song does not resume where it left off.

when quitting the app, maybe it should send a pause command

tried doing 

```rust
 match command {
        Command::Quit => {
            client_pub.send(ClientRequest::Player(PlayerRequest::Pause))?;
            ui.is_running = false;
        }
```

but didn't work, i don't know if there's a more correct way of doing it tho.

mvanhorn:
## Summary

Sends a Pause request when the user quits the TUI so playback stops instead of continuing in the background.

## Why this matters

If a song is playing and you close the app, it keeps playing through Spotify's servers. Reopening the app later doesn't resume where you left off. Pausing on quit fixes both problems.

## Changes

- `spotify_player/src/event/mod.rs`: Add `PlayerRequest::Pause` send in the `Command::Quit` handler, before setting `is_running = false`.

The `PlayerRequest::Pause` handler (in `client/mod.rs:296`) already checks `playback.is_playing`, so this is a no-op when nothing is playing. The flume channel `send()` queues the message synchronously, and the client consumer processes it during the UI loop's refresh sleep before `process::exit` runs.

## Testing

Verified the change compiles and follows the same pattern as other player commands (NextTrack, PreviousTrack, ResumePause) which all use `client_pub.send(ClientRequest::Player(...))`.

Fixes #944

This contribution was developed with AI assistance (Claude Code).

### DIFF
--- a/spotify_player/src/event/mod.rs
+++ b/spotify_player/src/event/mod.rs
@@ -569,6 +569,7 @@ fn handle_global_command(
 ) -> Result<bool> {
     match command {
         Command::Quit => {
+            client_pub.send(ClientRequest::Player(PlayerRequest::Pause))?;
             ui.is_running = false;
         }
         Command::NextTrack => {


mielpeeters:
Does this mean that playback will be paused when you're listening on another device through Spotify Connect?

I often use the TUI to start playing some playlist on a set of speakers, and I'd like that to continue when i close the TUI.

mvanhorn:
@mielpeeters Good catch -- yes, `PlayerRequest::Pause` hits Spotify's active-device endpoint, so if your active playback context is a Spotify Connect device (speakers, phone, web player), this will pause that too. The TUI doesn't distinguish between "playing through the TUI's own audio sink" vs "controlling a remote device" -- to Spotify they're the same playback session.

For your workflow (start speakers from the TUI, then close it) this PR is the wrong default. I'll gate it behind a config option (something like `pause_on_quit = false` with the default preserving current behavior) and update the PR. Thanks for flagging it before this landed.

PR #993, opened 2026-05-19
Added MPRIS OpenUri support

Bettehem:
Implemented MPRIS OpenUri and it now works with Spotify URIs, for example: 
`spotify:track:6CfwYtEnSDgMxosLZ5Vbiu`
and
`spotify:album:3HNzOyPbz5vPvUie7lI97X`

I'm still a beginner at rust so please let me know if there is anything that should be done differently


spotify_player/src/media_control.rs
```diff
@@ -127,11 +132,45 @@ pub fn start_event_watcher(
...
+                        Some(Playback::URIs(vec![episode_id], None))
+                    }
+                    _ => None,
+                };
+                if let Some(s) = playback {
+                    client_pub
+                        .send(ClientRequest::Player(PlayerRequest::StartPlayback(s, None)))
+                        .unwrap_or_default();
+                }
+            }
```
mielpeeters, line 175: I would implement it like this:

```suggestion
            MediaControlEvent::OpenUri(uri) => {
                let mut split = uri.split(':');
                let (Some("spotify"), Some(uri_type), Some(id)) =
                    (split.next(), split.next(), split.next())
                else {
                    return;
                };
                let id = id.to_string();
                let playback = match uri_type {
                    "album" => AlbumId::from_id(id)
                        .ok()
                        .map(|album_id| Playback::Context(ContextId::Album(album_id), None)),
                    "track" => TrackId::from_id(id)
                        .ok()
                        .map(|track_id| Playback::URIs(vec![PlayableId::Track(track_id)], None)),
                    "playlist" => PlaylistId::from_id(id).ok().map(|playlist_id| {
                        Playback::Context(ContextId::Playlist(playlist_id), None)
                    }),
                    "show" => ShowId::from_id(id)
                        .ok()
                        .map(|show_id| Playback::Context(ContextId::Show(show_id), None)),
                    "episode" => EpisodeId::from_id(id).ok().map(|episode_id| {
                        Playback::URIs(vec![PlayableId::Episode(episode_id)], None)
                    }),
                    _ => None,
                };
                if let Some(playback) = playback {
                    client_pub
                        .send(ClientRequest::Player(PlayerRequest::StartPlayback(
                            playback, None,
                        )))
                        .unwrap_or_default();
                }
            }

```

I'd argue that especially the parsing of the URI is more readable this way, plus it checks that the first part is actually 'spotify'.

The changes in the match statement get rid of the `unwrap()` calls you had, which is more safe.

Bettehem:
Thanks for the suggestion! I have now applied the changes

aome510 (approved):

PR #1005, opened 2026-06-15
feat: add `pause_on_startup` option to suppress startup auto-resume

### LINKED ISSUE #1004: Add `pause_on_startup` option to stop auto-resuming the previous session on startup

No. I went through the [Features README page](https://github.com/aome510/spotify-player?tab=readme-ov-file#features) and there is no option controlling startup playback state. The existing `autoplay` device option is unrelated (it only fetches radio tracks when a queue is exhausted).

When `spotify_player` is quit while a track is playing (e.g. `Ctrl+C` / `q`) and then reopened, it automatically resumes playing the last track. This is undesirable for users who want a quiet startup and just want to open the player without it immediately blasting audio.

The auto-resume is not initiated by `spotify_player` itself: on startup `initialize_playback` transfers playback with `play=false`. The resume comes from Spotify's server-side Connect state, which is restored by the integrated `librespot` client when it re-registers as a device. Because the trigger is server-side, there is currently no way to opt out from the client.

The only workaround today is the manual habit of pausing before quitting, which is easy to forget and not a real fix.

A config option, e.g. `pause_on_startup` (default `false`, opt-in, non-breaking), that pauses the playback Spotify auto-resumes on startup. When enabled, the integrated client pauses the first auto-started playback once, so the app opens quietly. Default `false` preserves the current resume-on-open behaviour for everyone who relies on it.

- Pausing before every quit (manual, error-prone, not automatic).
- `enable_streaming = "Never"` to stop the integrated client resuming, but that disables local streaming entirely, which defeats the purpose.

**Additional context**

I have a working implementation and will open a PR shortly. Behaviour summary:
- Opt-in via `pause_on_startup` (requires the `streaming` feature), default `false`.
- Scoped to the first streaming connection of the process, so `RestartIntegratedClient` mid-session is unaffected.
- No-op when nothing auto-resumes.

ecschoye:
## Summary

Adds an opt-in `pause_on_startup` config option (requires the `streaming` feature, default `false`) that stops `spotify_player` from auto-resuming the previous session when it starts.

Closes #1004.

## Problem

When the app is quit while playing and reopened, it immediately resumes the last track. The app itself does not start playback (`initialize_playback` transfers with `play=false`); the resume comes from Spotify's server-side Connect state, which the integrated `librespot` client restores when it re-registers as a device. There was no client-side way to opt out, so the only workaround was the manual habit of pausing before quitting.

## Solution

When `pause_on_startup = true`, the integrated client pauses the first playback that Spotify auto-resumes on startup:

- A process-level `IS_FIRST_CONNECTION` guard scopes the behaviour to the **first** streaming connection, so `RestartIntegratedClient` reconnecting mid-session does not pause an intentionally playing track.
- In the player-event loop, the first `Playing` event arms a one-shot `spirc.pause()` via the new `AppClient::pause_streaming_on_startup`. If the streaming connection is not registered yet, it stays armed and retries on the next `Playing` event.
- No-op when nothing auto-resumes.

Default is `false`, so existing resume-on-open behaviour is unchanged for everyone who relies on it.

## Changes

- `config/mod.rs`: add `pause_on_startup` field (+ default `false`), `streaming`-gated.
- `client/mod.rs`: add `AppClient::pause_streaming_on_startup`.
- `streaming.rs`: first-connection guard + one-shot pause in the player-event loop.
- `docs/config.md`: document the option.

## Testing

- `cargo build` (default features) — passes.
- `cargo clippy` — no warnings.
- `cargo fmt --check` — clean.
- `cargo check --no-default-features` — passes (confirms `cfg(streaming)` gating).
- Manual: with `pause_on_startup = true`, quitting while playing and reopening now starts paused; with it unset/`false`, the previous resume-on-open behaviour is unchanged.


ecschoye:
Implementation note on avoiding an audible startup blip:

The pause is triggered on the librespot **`Loading`** player event, not `Playing`. `Loading` fires before the audio sink starts, and `spirc`'s `handle_pause` transitions `LoadingPlay -> LoadingPause` (calling `player.pause()`), so the track finishes loading already paused and nothing is ever sent to the sink. A reactive pause on `Playing` would let a few milliseconds of audio through before stopping. `Playing` is kept only as a fallback in case `Loading` is skipped.

ecschoye:
Correction to my note above: pausing only on `Loading` is not reliable on its own. The librespot Connect transfer finalizes the play state asynchronously, so at the `Loading` event the spirc may not be in `LoadingPlay` yet, making the pause a no-op. The implementation now does:

- **`Loading`** -> best-effort `pause()` to avoid the audio blip; does **not** disarm (cannot confirm it took effect).
- **`Playing`** -> authoritative `pause()` (transfer has finalized), then disarm.
- **`Paused`** -> disarm only (the `Loading` pause took effect and the track loaded already paused, no audio).

This guarantees the session ends paused (the `Playing` arm is the backstop) while remaining blip-free whenever the early `Loading` pause wins.

aome510 (approved):

Issue #710, opened 2025-05-03
playback stops intermittently


earhax:
**Describe the bug**
intermittently, playback just stops in the middle of a song, even with the Spotify app open in addition to Spotify-player CLI.

**To Reproduce**
use Spotify-player, play music in a playlist. after a few songs, playback randomly just stops dead. Happens to me all the time. but maybe i'm just lucky like that!

**Expected behaviour**
playing songs shouldn't just stop for no reason. playback should continue until stopped by the user. 

**Log and backtrace**
Log and backtrace of a run reproducing the bug to help debug the issue, which are usually located in `$HOME/.cache/spotify-player/`.
i'm on my phone right now and don't have access to the log, but if this keeps up, I'll add the log later. also, i think $home only works on Mac. on PC it's probably %userprofile% and \ instead of / for the separator. might be worth noting in bug report steps/info.

**Environment**

- OS: windows 10 pro
- Application version: current/latest as of this report

stevenwalton:
This has been happening to me a lot lately. I was follow the log waiting for it to crash. [Here's the tail](https://0x0.st/8JU-.log). 
<sub><sup>Note that this has a max lifetime of a week and I redacted a few things like the `access_token`</sup></sub>

### Comments
It appears what is happening is that the server connection gets closed and after re-establishing the audio doesn't continue. 
***Sometimes*** I can continue by pressing play, but this doesn't always seem to be the case.
The log will say that it got a Paused pay event.
But most times when I press play the progress bar resumes but there is no audio (@earhax do you see this behavior?). 
Because of this, I think the issue isn't just due to resolving the connection.
I think there's something a bit more complicated and may involve the connection to the audio device.

I notice some weird behavior in the logs when it is re-establishing connections. 

```python
# This log has the crash
$ cat ~/.cache/spotify-player/spotify-player-25-05-08-14-38.log \
    | grep "expires_in" \
    | sed -e 's/\([^Z]*\)[^,]*, [^}]*}, \([^,]*\).*/\1Z \2/g'
2025-05-08T21:38:31.306186Z expires_at: Some(2025-05-08T22:38:31.306172Z)
2025-05-08T22:41:43.967958Z expires_at: Some(2025-05-08T23:41:43.967954Z)
2025-05-08T23:42:04.108905Z expires_at: Some(2025-05-09T00:42:04.108904Z)
2025-05-09T00:25:38.313487Z expires_at: Some(2025-05-09T01:25:38.313481Z)

# My active log
[...]

Issue #705, opened 2025-04-05
spotify-player device disappears after some time


contonice:
**Describe the bug**
I'm trying to use this app as a backend for a smart speaker.
My setup: raspberry pi with wired internet, usb sound card and speakers.
I am running the application in daemon mode using -d flag and systemd. Then, when I want to listen to music, I open spotify on my phone and use spotify connect to turn on the track on raspberry. This works immediately after running the script. But if a certain amount of time passes (one night for example), the spotify-player device will disappear. However, the application does not die, the process remains run in the system.

**To Reproduce**
1) Install an application with a flag to run as a daemon.
2) Change the ID in the config to your application created in the spotify api.
3) Launch the application with the -d flag.
4) Launch spotify on your phone and make sure that there is a spotify player in the list.
5) Wait a few hours.
6) Launch spotify on your phone again and see that the spotify player has disappeared.

Optional:
7) Kill the player process and launch it again (or use systemctl restart).
8) Open spotify on your phone.
9) spotify-player is working again.

**Expected behaviour**
As long as application working, there is spotify-player on phone.

**Log and backtrace**
Backtrace is empty.
Here is log:
https://drive.google.com/file/d/1AnGfBljq5IlkNGm4FwOR0mpmsixZhcmY/view

**Environment**

- OS: Latest Raspberry Pi OS (`Linux rpi 6.6.74+rpt-rpi-2712 #1 SMP PREEMPT Debian 1:6.6.74-1+rpt1 (2025-01-27) aarch64 GNU/Linux`)
- Application version: `0.20.4`
- Application features: `cargo install spotify_player --features daemon`

**Additional context**

My systemctl config:
```
[Unit]
Description=Spotify Player daemon
After=network.target

[Service]
Type=forking
ExecStart=/home/contie/.cargo/bin/spotify_player -d
Restart=on-failure

[Install]
WantedBy=default.target
```

aome510:
> Another solution is to disable the streaming on spotify-player and use spotifyd instead. Spotifyd never dies idle and has a service for systemd.
You can use spotify-player using spotifyd as a device but it's not necessary but you'll likely need a spotify connect API for spotify-player (to make the device showing on the list of devices).

Weird cause `spotifyd` also uses `librespot` under the hood. 

Looking into the logs, this seems to be the problem
```
2025-04-04T15:25:22.921036Z ERROR librespot_connect::spirc: ContextError: Error { kind: Unknown, error: StatusCode(502) }    
```
[...]

PR #1007, opened 2026-06-19
fix: fall back to an available device when starting playback with no active device

ecschoye:
## Summary

When the user starts a new playback (e.g. pressing Enter on a search result) and there is **no active playback at that moment**, `spotify_player` silently fails with "No playback found. Please start a new playback. Make sure there is a running Spotify device ...", even though the integrated `streaming` device is still available. Restarting the app fixes it. This makes `StartPlayback` recover the same way startup already does.

Related: #705, #710.

## Root cause

`PlayerRequest::StartPlayback` derives its target device solely from the current buffered playback:

```rust
let device_id = playback.as_ref().and_then(|p| p.device_id.as_deref());
self.start_playback(p, device_id).await?;
```

If the integrated device goes idle and Spotify drops it as the *active* device, the playback poller sets `player.playback = None` (and clears `buffered_playback`). After that, `device_id` is `None`, so `start_playback` sends no device id and the Web API targets "the currently active device", of which there is none, producing the error.

`initialize_playback` does not hit this because on startup it falls back to `find_available_device()`, which explicitly appends the integrated device id (`session.device_id()`) under the `streaming` feature. The runtime play path had no equivalent recovery, which is why only a restart helped.

## Fix

In the `StartPlayback` branch, when no device id is available from the current playback, fall back to `find_available_device()`, the exact recovery `initialize_playback` already uses:

```rust
let device_id = match playback.as_ref().and_then(|p| p.device_id.clone()) {
    Some(id) => Some(id),
    None => self.find_available_device().await?,
};
self.start_playback(p, device_id.as_deref()).await?;
```

Scoped to `StartPlayback` only. The other requests (`Resume`, `NextTrack`, `Pause`, ...) legitimately operate on an existing playback and keep their `no playback found` guard.

## Scope / limitation

This recovers the common case where the integrated device is still registered but no longer the *active* device. It does not address librespot fully deregistering the device after a network blip (the librespot-side reconnection discussed in #705 / #710), which is a separate concern.

## Testing

- `cargo build -p spotify_player` — passes.
- `cargo clippy -p spotify_player` — no warnings.
- `cargo check -p spotify_player --features streaming` — passes (exercises the integrated-device branch in `find_available_device`).

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -268,12 +268,19 @@ impl AppClient {
                 if let (Some(shuffle), Some(playback)) = (shuffle, playback.as_mut()) {
                     playback.shuffle_state = shuffle;
                 }
-                let device_id = playback.as_ref().and_then(|p| p.device_id.as_deref());
-                self.start_playback(p, device_id).await?;
+                // Resolve the target device. When there is no active playback (e.g. the integrated
+                // device went idle and Spotify dropped it as the active device), fall back to
+                // `find_available_device`, the same recovery `initialize_playback` uses on startup,
+                // so a new playback can be started without restarting the app.
+                let device_id = match playback.as_ref().and_then(|p| p.device_id.clone()) {
+                    Some(id) => Some(id),
+                    None => self.find_available_device().await?,
+                };
+                self.start_playback(p, device_id.as_deref()).await?;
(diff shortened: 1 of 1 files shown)


aome510:
not sure if it's a good idea if start playback triggers a playback on unexpected active device. The startup path only does connect so user can switch to different device if needed

Issue #899, opened 2025-12-25
App freezes when current playing song loads


MonkyMars:
**Describe the bug**
The app freezes completely. keystrokes do nothing. it also just stops listening to resizing.

**To Reproduce**
Open the app while having a song playing on another device

**Expected behaviour**
Should not freeze

**Log and backtrace**
2025-12-25T15:19:22.927352Z  INFO librespot_core::session: Country: "NL"
2025-12-25T15:19:23.039595Z  INFO spotify_player::streaming: New streaming connection has been established!
2025-12-25T15:19:23.039605Z  INFO spotify_player::client: Used a new session for Spotify client.
2025-12-25T15:19:23.134950Z  INFO spotify_player: Starting a client socket at 127.0.0.1:8080
2025-12-25T15:19:23.135120Z  INFO spotify_player::media_control: Initializing application's media control event watcher...
2025-12-25T15:19:23.135472Z  INFO librespot_core::spclient: Resolved "gew4-spclient.spotify.com:443" as spclient access point
2025-12-25T15:19:23.262699Z  INFO client_request{request=GetCurrentUser}: spotify_player::client: Successfully handled the client request, took: 127ms
2025-12-25T15:19:23.277223Z  INFO librespot_connect::spirc: active device is <3d077414f18724c02fc88768ccfadd0af10eb37f> with session <4gRLxP0KiqIPmdW5MLy0gk>
2025-12-25T15:19:23.364871Z  INFO client_request{request=GetUserPlaylists}: spotify_player::client: Successfully handled the client request, took: 229ms
2025-12-25T15:19:23.498515Z  INFO client_request{request=GetUserSavedShows}: spotify_player::client: Successfully handled the client request, took: 363ms
2025-12-25T15:19:23.582951Z  INFO client_request{request=GetUserSavedAlbums}: spotify_player::client: Successfully handled the client request, took: 447ms
2025-12-25T15:19:23.628137Z  INFO client_request{request=GetUserFollowedArtists}: spotify_player::client: Successfully handled the client request, took: 493ms

**Screenshots**

**Environment**
build from latest source on cachyos with niri and kitty.

MonkyMars:
it may be a niri or kitty issue because of the resizing. this hasnt happened anywhere else though.

chiragverma11:
[...]

PR #969, opened 2026-04-08
Migration to ratatui_image instead of viuer

sakashimaa:
Resolves #963 
Resolves #1003
Resolves #1001

Screenshot:
image

## Summary
- Replace `viuer` with `ratatui-image` for rendering cover images
- Remove manual `set_skip` / `clear_area` hacks — `ratatui-image` renders as a native ratatui widget
- Remove `viuer` initialization in `main.rs`, use `Picker::from_query_stdio()` for protocol auto-detection
- Remove `cover_img_scale` usage (no longer needed — scaling is handled by `ratatui-image`)

## Notes
`Picker::from_query_stdio()` must run after `enable_raw_mode()` but before `EnterAlternateScreen` for correct protocol detection.

In nested terminals (e.g. neovim floating terminal), stdio queries don't reach the actual terminal emulator, so the protocol falls back to Halfblocks. This is a known limitation of `ratatui-image`'s detection approach. Regular terminals (foot, kitty, iTerm2, etc.) work correctly

PR #1008, opened 2026-06-25
fix: prevent stdin race between viuer and event handler (#1003)

### LINKED ISSUE #1003: Interface won't load
**Describe the bug**
TUI never loads on startup, requires a seemingly random number of restarts to work. Sometimes I only need to restart once, sometimes 5-6 times before it loads correctly. Music will play normally regardless of whether the UI loaded or not.

**To Reproduce**
Open spotify-player via kitty (or konsole, it doesn't matter)
Interface doesn't load
Restart several times until it loads correctly

**Expected behaviour**
Open and display TUI normally.

**Log and backtrace**
[spotify-player-26-06-07-18-3.tar.gz](https://github.com/user-attachments/files/28688110/spotify-player-26-06-07-18-3.tar.gz)

**Screenshots**
https://github.com/user-attachments/assets/f2ab1fd7-748a-4d84-8bdc-4d6801db0d83

**Environment**

- OS: CachyOS
- Application version: v0.23.0
- Application features: cargo install spotify_player --features image

**Additional context**
I'm only experiencing this on Cachy (Arch based distro). Previously when I was using Nobara (Fedora based distro) the app worked perfectly fine.

Douglas019BR:
Fixes #1003 — TUI intermittently fails to load on startup when the image feature is enabled on Kitty (and potentially other terminals that support the Kitty graphics protocol).
  
  Root Cause :
  
  On startup, viuer::get_kitty_support() sends escape sequences to the terminal and reads
  responses from stdin to detect graphics protocol support. Previously, this ran in start_app()
   while the terminal event handler thread was spawned independently shortly after.
  
  The event handler calls crossterm::event::read() in a loop, which also reads from stdin. If
  the terminal was slow to respond to viuer's protocol query, the response bytes would arrive
  after the event handler started — causing crossterm to consume them. This left viuer (and
  subsequently the UI) in a broken state.
  
  Because the timing depends on terminal response speed, the bug is intermittent — sometimes it
  works on the first try, sometimes it takes 5-6 restarts.
  
  Fix:
  
  Two changes:
  
  1. Moved detect_image_protocol() into the UI thread — it now runs inside ui::run(), after
  init_ui() has already enabled raw mode. This ensures the terminal query happens in a
  controlled, single-reader context.
  2. Added a std::sync::Barrier — the event handler thread waits on the barrier before starting
  crossterm::event::read(). The UI thread releases the barrier only after
  detect_image_protocol() completes. This guarantees no concurrent stdin readers during the
  protocol detection.
  
  Execution order (before → after)
  
  Before (racy):
  
  main thread:  viuer::get_kitty_support() → spawn threads
  event thread: crossterm::event::read()  ← could steal viuer's response
  ui thread:    init_ui() → render loop
  
  After (synchronized):
  
  ui thread:    init_ui() → detect_image_protocol() → barrier.wait()
  event thread: barrier.wait() → crossterm::event::read()
  
  Related
  
  - This is the same class of issue as #899, where the workaround was downgrading viuer to
  v0.9.2. This fix addresses the underlying race condition rather than relying on
  version-specific behavior.
  - The Cargo.toml comment about needing to resolve the freezing issue before upgrading viuer
  should now be unblocked by this fix.
  
  ─────────────────────────────────────────────────────────────────────────────────────────────
  
  │ Note: I'm not a seasoned Rust developer, so I'm not entirely sure this is the most
  performant or idiomatic way to achieve this synchronization. If there's a lighter-weight
  primitive or a better pattern for this use case, I'm happy to adjust. The fix works reliably
  in my testing on Ubuntu + Kitty.

### DIFF
--- a/spotify_player/src/main.rs
+++ b/spotify_player/src/main.rs
@@ -97,17 +97,6 @@ fn init_logging(
 
 #[tokio::main]
 async fn start_app(state: &state::SharedState) -> Result<()> {
-    if !state.is_daemon {
-        #[cfg(feature = "image")]
-        {
-            // initialize `viuer` supports for kitty, iterm2, and sixel
-            viuer::get_kitty_support();
-            viuer::is_iterm_supported();
-            #[cfg(feature = "sixel")]
-            viuer::is_sixel_supported();
-        }
-    }
-
     // client channels
     let (client_pub, client_sub) = flume::unbounded::<client::ClientRequest>();
 
@@ -179,21 +168,23 @@ async fn start_app(state: &state::SharedState) -> Result<()> {
         })?;
 
     if !state.is_daemon {
-        // terminal event handler task
+        let terminal_ready = std::sync::Arc::new(std::sync::Barrier::new(2));
+
         std::thread::Builder::new()
             .name("terminal-event-handler".to_string())
             .spawn({
                 let client_pub = client_pub.clone();
                 let state = state.clone();
+                let terminal_ready = terminal_ready.clone();
                 move || {
+                    terminal_ready.wait();
                     event::start_event_handler(&state, &client_pub);
                 }
             })?;
 
-        // application UI task
         std::thread::Builder::new().name("ui".to_string()).spawn({
             let state = state.clone();
-            move || ui::run(&state)
+            move || ui::run(&state, terminal_ready)
         })?;
     }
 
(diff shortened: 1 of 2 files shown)


aome510:
Should be fixed by https://github.com/aome510/spotify-player/pull/969. Can you test that PR instead?

PR #1025, opened 2026-07-14
include compilations on artist album page

xdBronch:
f.e. on https://open.spotify.com/artist/2jl4qd6UbzeCmImT4nWbtA compilation now shows
image


aome510 (commented):
don't think it's a good idea to include compilation since specific artists will require polling A LOT of albums. If you want this to be included, better to make it configurable

PR #1047, opened 2026-07-28
media_control: add xesam:url to MPRIS metadata

Bettehem:
souvlaki v0.9 is under development and has a bunch of breaking changes from v0.8

However, it seems the development of both 0.8 and 0.9 has been inactive for quite a while, so I thought I'd fork the repo and start adding missing features.

Instead of trying to make the 0.9 branch work with the possibility of more breaking changes being introduced, I forked the stable 0.8 master branch and added the missing xesam:url metadata field to it.

I would like to use spotify-player but I control my music players using an external program and xesam:url is one important field missing from the metadata which would be needed for it.

Specification: https://www.freedesktop.org/wiki/Specifications/mpris-spec/metadata/#xesam:url

Link to the xesam:url addition commit which is used my fork of souvlaki: https://github.com/Bettehem/souvlaki/commit/17083b6a8324d27ffea6b25f03a90b9e3974ff85

EDIT: added link to the souvlaki commit used in this PR

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5746,8 +5746,7 @@ dependencies = [
 [[package]]
 name = "souvlaki"
 version = "0.8.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5855c8f31521af07d896b852eaa9eca974ddd3211fc2ae292e58dda8eb129bc8"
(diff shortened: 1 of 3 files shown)


aome510 (commented):
we cannot publish a new version on crates.io with git package. Can you upstream the changes instead?

Bettehem:
Yes, I will try!

Issue #974, opened 2026-04-12
Spotify player perpetually loads and never opens


RGH271:
**Describe the bug**
When I open spotify player on my mac, it doesn't load. Instead it shows the outline of the windows and a loading spinner, and doesn't load (at least not in a reasonable amount of time)

**To Reproduce**
run `spotify_player` in terminal

**Expected behaviour**
Spotify_player to load

**Log and backtrace**
Log and backtrace of a run reproducing the bug to help debug the issue, which are usually located in `$HOME/.cache/spotify-player/`.

Please refer to [this URL](https://github.com/aome510/spotify-player?tab=readme-ov-file#logging) for more details.

[...]

Issue #890, opened 2025-12-22
Failed to handle client request: status code 429 Too Many Requests


DavidHospital:
**Describe the bug**
`spotify-player` seemingly cannot make any requests to Spotify it seems. Any action that requires a request to spotify fails silently and the TUI is stuck loading forever.

**To Reproduce**
I am not sure to be honest, this just started happening every time I try to use `spotify-player` starting today. I have double checked there aren't any extra instancing of it running in the background.

**Expected behaviour**
I expect the basic functions of the app to work.

**Log and backtrace**
An example of some of the logs:
```
2025-12-22T18:26:42.112431Z  INFO client_request{request=GetContext(Album(AlbumId("6DxidEI7EhF8md8ev83iNY")))}: spotify_player::client: Get album context: spotify:album:6DxidEI7EhF8md8ev83iNY
2025-12-22T18:26:42.176631Z ERROR client_request{request=GetContext(Album(AlbumId("6DxidEI7EhF8md8ev83iNY")))}: spotify_player::client::handlers: Failed to handle client request: http error: status code 429 Too Many Requests
```
It is the same 429 error for any action.

**Environment**

- OS: Ubuntu 22.04.5 LTS x86_64
- Application version: 0.21.2
- Application features: default features (ie. `cargo install spotify_player`)

**Additional context**
This began happening this morning (around 10am EST) when I loaded it up to play some music. I've tried it again periodically throughout the day with no luck. I've also double checked my other devices to see if, for example, I could use Spotify on my phone; no issues there which rules out a problem with my account, etc.

aome510:
I noticed the same thing when coming back to this project and start working on it again today (bad timings 😢 ). 

The rate-limit issue happens when making API requests to the official Spotify client (client ID: `65b708073fc0480ea92a077233ca87bd`). Hopefully, this is just a temporary thing

aome510:
In case anyone wants a temporary workaround, I got this app working using user-provided client instead of the integrated client to make API calls in https://github.com/aome510/spotify-player/pull/891. 

With the above change, API requests will be sent to **user-provided client instead of the official Spotify client**. One limitation of this workaround is that with user-provided client, `spotify_player` will have a limited API access compared to the full access when using the official Spotify client (see https://developer.spotify.com/blog/2024-11-27-changes-to-the-web-api for more details)

[...]

PR #1048, opened 2026-07-29
fix(client): serialize search to reduce 429 QUOTA_EXCEEDED

louzt:
## Summary

This PR targets the most common cause of `429 Too Many Requests / QUOTA_EXCEEDED` errors affecting the `AppClient::search()` path. Symptoms include empty search results, dropped queries mid-scroll, and intermittently failing mixed-type searches — irrespective of the front-end (TUI, daemon consumer, CLI, or third-party shell integration).

Two underlying issues compound:

1. `search()` fires six parallel requests (one per `SearchType`) under `tokio::try_join!`. Spotify rate-limits this burst even for clients on the extended-quota mode (#890, #974).
2. The mixed-results endpoint does not always populate the fields rspotify's typed models require (e.g. `followers` on artists), so the Show / Episode branches fail to deserialize and make the entire `try_join!` return an error — turning a partial-success case into a full failure.

This is related to, but complementary with, the NCSPOT extended-quota fallback in #891.

## Changes

- **`client/mod.rs::search()`** — drop Show / Episode searches and serialize the four remaining requests in order. Reduces request burst from 6 to 4, eliminates the deserialization-induced full-failure mode for mixed queries, and removes the parallel burst that triggers rate-limiting.
- `SearchResults { tracks, artists, albums, playlists, shows, episodes }` keeps the same shape; `shows` and `episodes` are now always empty vectors.

## Validation

```sh
cargo fmt --all -- --check
cargo check --no-default-features --features rodio-backend,media-control,streaming
cargo clippy --no-default-features --features rodio-backend,media-control,streaming -- -D warnings
```

All three pass clean.

## Repro / environment

Verified by the reporter on:

- **spotify-player**: master at `v0.24.1-1-g51cb762`, Cargo.toml version `0.24.1`
- **OS**: Debian forky/sid (kernel `6.19.14+deb14-amd64`)
- **Compositor**: niri (Wayland)
- **Terminal**: Ghostty (`TERM_PROGRAM=ghostty`)
- **Run command**: `cargo build --release --features streaming,image,notify,media-control,rodio-backend,fzf` then launched from a `distrobox enter arch-loust` Arch container
- **Client ID**: bundled default (`NCSPOT_CLIENT_ID` constant); custom IDs pre-2024-11-27 are not required to repro

## Notes

- The NCSPOT fallback in #891 remains the primary mitigation for users with a custom `client_id`; this PR is complementary.
- A short-TTL cache on top of `search()` would remove the remaining repeat-query pressure; happy to send that as a separate PR if useful.

### DIFF
--- a/spotify_player/src/client/mod.rs
+++ b/spotify_player/src/client/mod.rs
@@ -1080,72 +1080,57 @@ impl AppClient {
 
     /// Search for items (tracks, artists, albums, playlists) matching a given query
     pub async fn search(&self, query: &str) -> Result<SearchResults> {
-        let (
-            track_result,
-            artist_result,
-            album_result,
-            playlist_result,
-            show_result,
-            episode_result,
-        ) = tokio::try_join!(
-            self.search_specific_type(query, rspotify::model::SearchType::Track),
-            self.search_specific_type(query, rspotify::model::SearchType::Artist),
-            self.search_specific_type(query, rspotify::model::SearchType::Album),
-            self.search_specific_type(query, rspotify::model::SearchType::Playlist),
-            self.search_specific_type(query, rspotify::model::SearchType::Show),
-            self.search_specific_type(query, rspotify::model::SearchType::Episode)
-        )?;
+        // Show and Episode searches are skipped: the mixed-results endpoint does
+        // not always populate fields rspotify's typed models require (e.g. artist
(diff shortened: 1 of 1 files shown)


aome510:
Single search is implemented in https://github.com/aome510/spotify-player/pull/1077/changes/0210b81b17bba5d81f7cf3a6f1e9c2c8a945f212

PR #1070, opened 2026-08-28
feat: reload theme dynamically on config changes

Bogieloo:
### Summary
Adds live theme reloading to the UI loop without requiring an application restart.

### Changes
- Check `theme.toml` modification time periodically in the UI loop.
- Reload and update `ui.theme` in-place when the theme configuration changes.
- Allow user-defined themes in `theme.toml` to update existing theme definitions by name.


FidelSch:
Maybe we could avoid constant polling by using something like [notify](https://docs.rs/notify/latest/notify/)

